PR #12, opened 2026-02-03
Minor general fixes suggestions

filip-komarzyniec:
- introduced multiple minor fixes to the code -- mostly in case of different functions' parameters definition and type hinting, 

- introduced `isort` tool to the project; run the `isort` on all project files

- ~~highlighted a couple of tougher aspects up for reconsideration. They're marked in appropriate places as TODO comments. If deemed justified and necessary I'd suggest opening a separate task in order to tackle each problem.~~

### DIFF
--- a/ai4rag/core/experiment/experiment.py
+++ b/ai4rag/core/experiment/experiment.py
@@ -13,8 +13,8 @@ from llama_stack_client import LlamaStackClient
 from ai4rag import logger
 from ai4rag.core.experiment.benchmark_data import BenchmarkData
 from ai4rag.core.experiment.exception_handler import (
-    AssetSaveError,
     AI4RAGError,
+    AssetSaveError,
     ExperimentExceptionsHandler,
     IndexingError,
 )
@@ -28,7 +28,7 @@ from ai4rag.core.experiment.utils import (
     get_retrieval_params,
     query_rag,
 )
-from ai4rag.core.hpo.base_optimiser import BaseOptimiser, OptimiserSettings, OptimisationError
+from ai4rag.core.hpo.base_optimiser import BaseOptimiser, OptimisationError, OptimiserSettings
 from ai4rag.core.hpo.random_opt import FailedIterationError, RandomOptimiser
 from ai4rag.evaluator.base_evaluator import BaseEvaluator, EvaluationData, MetricType
 from ai4rag.evaluator.unitxt_evaluator import UnitxtEvaluator
@@ -37,17 +37,13 @@ from ai4rag.rag.embedding.base_model import EmbeddingModel
 from ai4rag.rag.foundation_models.base_model import FoundationModel
 from ai4rag.rag.retrieval.retriever import Retriever
 from ai4rag.rag.template.rag_template import LlamaStackRAG
+from ai4rag.rag.vector_store.get_vector_store import get_vector_store
 from ai4rag.search_space.src.models import EmbeddingModels
 from ai4rag.search_space.src.parameter import Parameter
 from ai4rag.search_space.src.search_space import AI4RAGSearchSpace
-from ai4rag.utils.constants import (
-    AI4RAGParamNames,
-    EventsToReport,
-    ExperimentStep,
-)
+from ai4rag.utils.constants import AI4RAGParamNames, EventsToReport, ExperimentStep
 from ai4rag.utils.event_handler.event_handler import BaseEventHandler, LogLevel
 from ai4rag.utils.experiment_monitor import ExperimentMonitor
-from ai4rag.rag.vector_store.get_vector_store import get_vector_store
 
 
 class AI4RAGExperiment:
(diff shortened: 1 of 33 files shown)


ai4rag/core/experiment/experiment.py
```diff
@@ -28,7 +28,7 @@
     get_retrieval_params,
     query_rag,
 )
-from ai4rag.core.hpo.base_optimiser import BaseOptimiser, OptimiserSettings, OptimisationError
+from ai4rag.core.hpo.base_optimiser import BaseOptimiser, OptimisationError, OptimiserSettings
```
jakub-walaszczyk, line 31: 👀 github statistics improvement :P?

ai4rag/core/experiment/experiment.py
```diff
@@ -547,7 +525,9 @@ def objective_function(space: dict) -> float | None:
                 raise FailedIterationError(msg) from err
 
         # MPS - models pre-selection based on sample evaluation. Run if there are more than 3 foundation models
-        foundation_models = list(self.search_space[AI4RAGParamNames.FOUNDATION_MODEL].values)
+        foundation_models = list(
+            self.search_space[AI4RAGParamNames.FOUNDATION_MODEL].values
+        )  # TODO the keys for FM an EM in search space are out-of-sync with what's actually returned from search space
```
jakub-walaszczyk: Did you fix that or simply highlighted that this is "To do"?

ai4rag/core/experiment/mps.py
```diff
@@ -21,19 +24,7 @@
...
-from ai4rag.core.experiment.benchmark_data import BenchmarkData
-from ai4rag.utils.constants import (
-    AI4RAGParamNames,
-    ExperimentStep,
-    EventsToReport,
-)
-
-from ai4rag import logger
-
+from ai4rag.utils.constants import AI4RAGParamNames, EventsToReport, ExperimentStep
```
jakub-walaszczyk, line 27: What are the enhancements here? Simply reordering imports for some particular reason?

ai4rag/core/experiment/mps.py
```diff
@@ -94,6 +85,7 @@ class ModelsPreSelector:
     def __init__(
         self,
         metric: str,
+        # TODO search space does not hold list[FoundationModel] but rather list[ai4rag.search_space.prepare.input_payload_types::AI4RAGModel]
```
jakub-walaszczyk: It does hold list of `FoundationModel` as far as I am concerned. Now you are passing instance of the `FoundationModel` or the instance of the class inheriting from that.

Please clarify on what you mean. You are most likely still using `prepare_ai4rag_search_space` but this is out of date and still under rafactoring. Please use `AI4RAGSearchSpace` directly. That should resolve majority of the mentioned issues.

ai4rag/core/experiment/experiment.py
jakub-walaszczyk: Please do not use `prepare_ai4rag_search_space`. This needs to be refactored. It's in the scope for the next PRs. Use `AI4RAGSearchSpace` constructor directly.

ai4rag/search_space/src/parameter.py
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # -----------------------------------------------------------------------------
 from dataclasses import dataclass, fields
-from typing import Any, Literal, Optional, Sequence, TypeVar, Protocol, Generic
+from typing import Any, Generic, Literal, Optional, Protocol, Sequence, TypeVar
```
jakub-walaszczyk, line 6: Is there a specific reason why you are doing it alphabetically? Is it PEP-recommended to order imports like that?

ai4rag/search_space/src/parameter.py
```diff
@@ -29,6 +29,7 @@ class Parameter(Generic[HashableType]):
     param_type: Literal["B", "I", "R", "C"]
     v_min: Optional[int | float] = None
     v_max: Optional[int | float] = None
+    # TODO in most cases this is a list of dicts (which are not hashable). To be thought-through once again
```
jakub-walaszczyk: Once again, please do not use `prepare_ai4rag_serach_space` and remove this comment as this is not valid comment here. Thanks!

jakub-walaszczyk (changes requested):
Please notice that `prepare_ai4rag_search_space` is in the queue for refactoring as it is not ready to use at this stage. Use direct call to `AI4RAGSearchSpace` as in the example under `dev_utils`.

Notice that you are raising PR to the main branch. Use `dev` instead. Also fix the DCO as requested.

ai4rag/core/experiment/experiment.py
filip-komarzyniec: I guess it's local `isort` doings. Should we include that into the package or discard at all, wdyt?

ai4rag/core/experiment/experiment.py
filip-komarzyniec: deleted TODO as it'll be taken care of during `prepare_ai4rag_search_space` function refactoring and adjusting to the new flow.

ai4rag/core/experiment/experiment.py
filip-komarzyniec: I did that as part of the latest commit. If we do not want to proceed than it's just removing the latest commit.

jakub-walaszczyk (approved):

PR #21, opened 2026-02-19
Introduce AutoRAG sample notebook & `publish-pypi.yml` workflow

witold-nowogorski:
#### AutoRAG notebook
This PR introduces sample notebook demonstrating AutoRAG experiment flow.
Notebook link: https://github.com/IBM/ai4rag/blob/dev-samples/samples/run_ai4rag.ipynb

#### `publish-pypi.yml` workflow
As we are going to release on test pypi and prod pypi, new workflow for that has been introduced.

## Motivation
Currently there is no such sample in the repo.
No workflow for publish existed so far.

## Changes
- run_ai4rag.ipynb added under new directory: `samples`
- workflow for publish added under `.github/workflows/.publish-pypi.yml`
- Parameter has now optional `param_type` as we usually use `C` anyway.

## Testing
I've run the notebook:)
Workflow has not been tested yet as it need to be on the repo first. Let's test it:)

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] All checks passing


LukaszCmielowski (changes requested):
@witold-nowogorski please use notebook template and align the layout. You can follow up AutoML Predictor Notebook: https://github.com/LukaszCmielowski/prototypes/blob/main/AutoML/notebooks/classification/automl_predictor_notebook.ipynb

LukaszCmielowski (changes requested):
1. Let's add some | tail command to pip install to not log screens of outputs
2. For note let's introduce emoji/icon please see tip / action in the AutoML notebook for reference
3. Let's add some description what is required and why instead of 
> Set bucket name and environment variables to connect to S3 instance
4. Setting S3 env variable - there should be some action statement befoew so the user knows there is action to be performed. Add the IF to check if those are not available out-of-the-box ... should be on the cluster
5. Any reason why documents filenames are document_0.pdf instead of sth meaningful ?
6. Documents processing results - is a new section ? - if yes it is not part of table of contents ... From other hand there are some step 1, 2 enumerated. Let's remove that ans stick to markdown header syntax - if there are subsection just use less "#"

witold-nowogorski:
@LukaszCmielowski Comments resolved

LukaszCmielowski (approved):

PR #29, opened 2026-03-03
Penalty in hpo and optional embedding dimension (dynamic set)

jakub-walaszczyk:
## Description
Penalty for failing iterations added (0 returned instead of `None`) and adding dynamic embedding dimension discovery.

## Motivation
- penalty for better patterns recommendation
- embedding dimension dynamically added to simplify search space creation in KFP

## Changes
- Penalty for failing iterations
- dynamic embedding model dimension setting

## Testing
Run unit tests

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] All checks passing

### DIFF
--- a/ai4rag/core/hpo/gam_opt.py
+++ b/ai4rag/core/hpo/gam_opt.py
@@ -124,7 +124,7 @@ class GAMOptimizer(BaseOptimizer):
         for _ in range(iterations_limit):
             self._run_iteration()
 
-        successful_evaluations = [evaluation for evaluation in self.evaluations if evaluation["score"] != 0]
+        successful_evaluations = [evaluation for evaluation in self.evaluations if evaluation["score"] >= 0]
         if not successful_evaluations:
             raise OptimizationError("Number of evaluations has reached limit. All iterations have failed.")
 
@@ -157,7 +157,7 @@ class GAMOptimizer(BaseOptimizer):
         while successful_evaluations < self.settings.n_random_nodes:
             params = next(gen)
             score = self._objective_function(params=params)
-            if score >= 0:
+            if score > 0:
                 successful_evaluations += 1
             self._evaluated_combinations.append(params)
             params_with_score = params | {"score": score}
@@ -174,7 +174,6 @@ class GAMOptimizer(BaseOptimizer):
         """
         self._prepare_encoder()
         df = pd.DataFrame(data=self.evaluations)  # --> These are already known observations with scores.
-        df = df[df["score"].notna()].copy()
         data = df.drop(columns=["score"])
         target = df["score"]
 
@@ -253,7 +252,7 @@ class GAMOptimizer(BaseOptimizer):
 
         return remaining
 
-    def _objective_function(self, params: dict) -> float | None:
+    def _objective_function(self, params: dict) -> float:
         """
         Wrapper around the objective function provided to the optimizer.
 
@@ -264,9 +263,9 @@ class GAMOptimizer(BaseOptimizer):
 
         Returns
         -------
-        float | None
+        float
             Optimization score achieved for single node evaluation.
-            If None - iteration has ended up with a failed status.
+            Returns -1 if the iteration failed (used as a penalty sentinel).
         """
 
         try:
@@ -274,6 +273,6 @@ class GAMOptimizer(BaseOptimizer):
             loss = self.objective_function(params)
 
         except FailedIterationError:
-            loss = 0
+            loss = -1
 
         return loss
--- a/ai4rag/core/hpo/gam_opt.py
+++ b/ai4rag/core/hpo/gam_opt.py
@@ -170,7 +170,7 @@ class GAMOptimizer(BaseOptimizer):
         """
         Run single optimization iteration that consists of training GAM model
         to predict score for remaining nodes in the solutions space and choose
-        the best n ones to further evaluation.
+        the best n ones for further evaluation.
         """
         self._prepare_encoder()
         df = pd.DataFrame(data=self.evaluations)  # --> These are already known observations with scores.
--- a/ai4rag/core/hpo/gam_opt.py
+++ b/ai4rag/core/hpo/gam_opt.py
@@ -124,7 +124,7 @@ class GAMOptimizer(BaseOptimizer):
         for _ in range(iterations_limit):
             self._run_iteration()
 
-        successful_evaluations = [evaluation for evaluation in self.evaluations if evaluation["score"] is not None]
+        successful_evaluations = [evaluation for evaluation in self.evaluations if evaluation["score"] != 0]
         if not successful_evaluations:
             raise OptimizationError("Number of evaluations has reached limit. All iterations have failed.")
 
@@ -157,7 +157,7 @@ class GAMOptimizer(BaseOptimizer):
         while successful_evaluations < self.settings.n_random_nodes:
             params = next(gen)
             score = self._objective_function(params=params)
-            if score is not None:
+            if score >= 0:
                 successful_evaluations += 1
             self._evaluated_combinations.append(params)
             params_with_score = params | {"score": score}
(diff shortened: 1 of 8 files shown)


ai4rag/core/hpo/gam_opt.py
```diff
@@ -169,12 +169,11 @@ def evaluate_initial_random_nodes(self) -> None:
     def _run_iteration(self) -> None:
         """
         Run single optimization iteration that consists of training GAM model
-        to predict score for remaining nodes in the solutions space and chose
-        best n ones to further evaluation.
+        to predict score for remaining nodes in the solutions space and choose
+        the best n ones to further evaluation.
```
filip-komarzyniec: `the best n ones to further evaluation.`  --> `the best n ones FOR further evaluation.` would sound better, wouldn't it?

ai4rag/core/hpo/gam_opt.py
```diff
@@ -264,17 +263,16 @@ def _objective_function(self, params: dict) -> float | None:
 
         Returns
         -------
-        float | None
+        float
             Optimization score achieved for single node evaluation.
-            If None - iteration has ended up with a failed status.
+            Returns -1 if the iteration failed (used as a penalty sentinel).
```
filip-komarzyniec, line 268: Why did we decide to change `None` to `-1`? Was it proven anyhow (even empirically) that GAM performs better when predicting on sets containing failed calculations? 

I think the failures, most frequently, are not connected anyhow with the parameters themselves. They're rather due to the network issues, DB instances failures or internal bugs in used libraries. 
Taking it into account, what's the added value of including the failed iterations in the training set? Does it have to do anything with how GAM works specifically?

ai4rag/rag/embedding/llama_stack.py
```diff
@@ -28,21 +28,47 @@ class LSEmbeddingParams:
...
+        Note: This method is called during initialization when ``embedding_dimension``
+        is not explicitly provided.  It issues a real API request to the Llama Stack
+        server, so the server must be reachable at construction time.
+
+        Raises
+        ------
+        RuntimeError
+            When the embedding dimension cannot be determined (e.g. the server
+            is unreachable or the model is not available).
+        """
```
filip-komarzyniec, line 63: The docstring style throughout the KFP-related libraries seem to be the Google style. 

As we're currently developing ai4rag in parallel with multiple KFP-based libraries maybe we can think about switching the docstrings to Google style? 
From the dev POV it's naturally easier to stick to one approach.
I get the impression it's more common and popular than numpy one as well. 

Was there any specific requirement for the docstrings (or the style's feature) that once shifted the decision towards numpy style? 
Do you think it's sensible, beneficial or worth a shot at all to tingle with docstrings now?

ai4rag/rag/embedding/openai_model.py
```diff
@@ -15,6 +15,41 @@ class OpenAIEmbeddingModel(BaseEmbeddingModel):
     def __init__(self, client: OpenAI, model_id: str, params: dict[str, Any] | None = None):
         super().__init__(client=client, model_id=model_id, params=params)
 
+    @property
+    def params(self) -> dict[str, Any]:
+        return self._params
```
filip-komarzyniec, line 20: It seems there is an inconsistency of handling the embedding params between `LSEmbeddingModel` and `OpenAIEmbeddingModel`. 

The first scenario uses dedicated dataclass which enforces the specific attributes. Here, using just a dict, one can pass anything. 

Do we want to unify the approaches in the future?

ai4rag/rag/embedding/openai_model.py
```diff
@@ -15,6 +15,41 @@ class OpenAIEmbeddingModel(BaseEmbeddingModel):
...
+            self._params = params
+        if "embedding_dimension" not in self._params:
+            self._params["embedding_dimension"] = self._detect_embedding_dimension()
+
+    def _detect_embedding_dimension(self) -> int:
+        """Detect embedding dimension by making a minimal embedding call.
+
+        Note: This method is called during initialization when ``embedding_dimension``
+        is not present in the params dict.  It issues a real API request to the
+        OpenAI-compatible endpoint, so the service must be reachable at construction time.
```
filip-komarzyniec: technically speaking the endpoint should be OpenAI API compliant, not OpenAI-compatible. OpenAI is just a company's name 🙃

tests/ai4rag/rag/embedding/test_llama_stack.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class TestLSEmbeddingModel:
+    """Test suite for LSEmbeddingModel class."""
+
+    @pytest.fixture
+    def mock_ls_client(self, mocker):
+        """Create a mock Llama Stack client with default embedding response for auto-detection."""
+        mock_client = mocker.MagicMock()
+        mock_client.embeddings.create.return_value = _make_mock_ls_embedding_response(mocker, [[0.1, 0.2, 0.3, 0.4]])
```
filip-komarzyniec, line 29: This seems a bit overcomplicated. I did not tested the proposed solution, but wouldn't something like this substitute the `_make_mock` function at all? 

```python 
MagicMock(data=[MagicMock(embedding=emb) for emb in [[0.1, 0.2, 0.3, 0.4]]])

```

tests/ai4rag/rag/embedding/test_llama_stack.py
```diff
@@ -0,0 +1,134 @@
...
+        mock_data.embedding = emb
+        mock_response.data.append(mock_data)
+    return mock_response
+
+
+class TestLSEmbeddingModel:
+    """Test suite for LSEmbeddingModel class."""
+
+    @pytest.fixture
+    def mock_ls_client(self, mocker):
```
filip-komarzyniec, line 26: What is the benefit of defining the fixture inside the class?

filip-komarzyniec (approved):
/lgtm

🦾🧏🏼‍♂️🤫

ai4rag/core/hpo/gam_opt.py
jakub-walaszczyk: Ideally we focus on a single experiment and if one combination is failing (e.g. chunking settings causes exceeding context length of an embedding model) we want to highlight that this set of parameters will not work and it should not be recommended by the algorithm for next iterations. We assume that usage of the model like GAM actually allows us to find better correlation between parameters so ideally after some iterations model should "know" more about the dataset. 

Ideally, but with 4-8 iterations it does not really matter, right :D?

We do not have proper benchmarks yet, but this is to be extra cautious. It should not mess anything up tbh.

tests/ai4rag/rag/embedding/test_llama_stack.py
jakub-walaszczyk: I would say that if you do not want to use this fixture outside of the class scope it can be done like this. It provides no drawbacks and a tiny benefit, so we can keep that (Claude assisted on unit tests generation, I have reviewed these tests and they are legit)

ai4rag/rag/embedding/llama_stack.py
jakub-walaszczyk: @filip-komarzyniec legit proposal, can you open an issue for that in the `ai4rag` issues as an enhancement proposal? (GH issues, not jira)

ai4rag/core/hpo/gam_opt.py
jakub-walaszczyk: ✅

ai4rag/rag/embedding/openai_model.py
jakub-walaszczyk: In the future maybe, at this moment we did not investigate OpenAI API that much, we base on the LS mostly at the moment. Both classes accept dict so it is actually a bit unified. Closign as resolved rn

ai4rag/rag/embedding/openai_model.py
jakub-walaszczyk: ✅

tests/ai4rag/rag/embedding/test_llama_stack.py
jakub-walaszczyk: Maybe, but this works properly and is not hard to read. Let's revisit once tests for this module will be changed

PR #33, opened 2026-03-05
Release 0.3.1

jakub-walaszczyk:
Changes:
- Update tokens estimation constants
- Introduce `format.sh` script in scripts

### DIFF
--- a/ai4rag/search_space/src/search_space.py
+++ b/ai4rag/search_space/src/search_space.py
@@ -80,10 +80,9 @@ def _rule_chunk_size_within_embedding_context_length(combination: dict) -> bool:
         Whether the estimated token count fits within the embedding model's context length.
     """
     chunk_size = combination.get(AI4RAGParamNames.CHUNK_SIZE)
-    chunk_overlap = combination.get(AI4RAGParamNames.CHUNK_OVERLAP)
     embedding_model = combination.get(AI4RAGParamNames.EMBEDDING_MODEL)
 
-    if chunk_size is None or chunk_overlap is None or embedding_model is None:
+    if chunk_size is None or embedding_model is None:
         return True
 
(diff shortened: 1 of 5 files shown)


ai4rag/search_space/src/search_space.py
filip-komarzyniec, line 92: Can we unify the params to always be a `dict` or `dataclass` across all supported model types? 

Because it's inconsistent now I get the feeling it requires such unnecessary logic.

ai4rag/search_space/src/search_space.py
```diff
@@ -95,7 +95,7 @@ def _rule_chunk_size_within_embedding_context_length(combination: dict) -> bool:
     if context_length is None:
         return True
 
-    chars_per_token = 4.5
+    chars_per_token = 3.6
```
filip-komarzyniec, line 97: How can we ensure the overestimation will not happen again?

ai4rag/search_space/src/search_space.py
filip-komarzyniec: It would also be beneficial for the Model classes to return the `params` as dict type, not a `dataclass`.

filip-komarzyniec (approved):
🤫🧏🏼‍♂️

ai4rag/search_space/src/search_space.py
jakub-walaszczyk: the prvious value was underestimating tokens. The value 3.6 actually overestimates. The logic is reversed, and value should've been lowered comparing to original 4 chars per token in english dictionary.

Changing this value from 4 to lower actually will prevernt this behaviour. We will never be sure as we do not know the exact tokenizer user, but I think this is a fair idea to estimate.

ai4rag/search_space/src/search_space.py
jakub-walaszczyk: I will look at this idea in the next PRs

PR #65, opened 2026-05-19
Migrate from chat.completions to Responses API

filip-komarzyniec:
## Description
- change API calls from `chat/completions` to `responses` (both OpenAI-compatible)
- modified streamed pattern according to the document: https://github.com/LukaszCmielowski/architecture-decision-records/blob/autox_docs_updates/documentation/components/autorag/features/rag_pattern_inference.md
- other small changes to different classes and functions not really affecting the overall behaviour

## Motivation
- Use the built-in agent loop working under the hood of Responses API, 
- stay on top of the industry standards

## Changes
- API calls change
- small refactoring of various names so that they better represent current behaviour
- 

## Testing
Commit with updated tests will be added once manual tests are completed.

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/core/experiment/experiment.py
+++ b/ai4rag/core/experiment/experiment.py
@@ -28,7 +28,11 @@ from ai4rag.core.experiment.utils import (
     get_retrieval_params,
     query_rag,
 )
-from ai4rag.core.hpo.base_optimizer import BaseOptimizer, OptimizationError, OptimizerSettings
+from ai4rag.core.hpo.base_optimizer import (
+    BaseOptimizer,
+    OptimizationError,
+    OptimizerSettings,
+)
 from ai4rag.core.hpo.gam_opt import GAMOptimizer
 from ai4rag.core.hpo.random_opt import FailedIterationError
 from ai4rag.evaluator.base_evaluator import BaseEvaluator, EvaluationData, MetricType
@@ -154,7 +158,12 @@ class AI4RAGExperiment:
 
         self.job_id = kwargs.pop("job_id", "ai4rag_job_a0b1c2d3").replace("-", "_")
         self.metrics: Sequence[str] = kwargs.pop(
-            "metrics", (MetricType.ANSWER_CORRECTNESS, MetricType.FAITHFULNESS, MetricType.CONTEXT_CORRECTNESS)
+            "metrics",
+            (
+                MetricType.ANSWER_CORRECTNESS,
+                MetricType.FAITHFULNESS,
+                MetricType.CONTEXT_CORRECTNESS,
+            ),
         )
         self.evaluator: BaseEvaluator = kwargs.pop(
             "evaluator",
@@ -163,8 +172,12 @@ class AI4RAGExperiment:
         self.n_mps_foundation_models = kwargs.pop(
             "n_mps_foundation_models", ModelsPreSelector.DEFAULT_N_FOUNDATION_MODELS
         )
-        self.n_mps_embedding_models = kwargs.pop("n_mps_embedding_models", ModelsPreSelector.DEFAULT_N_EMBEDDING_MODELS)
-        self.known_observations: list[dict] | None = kwargs.pop("known_observations", None)
+        self.n_mps_embedding_models = kwargs.pop(
+            "n_mps_embedding_models", ModelsPreSelector.DEFAULT_N_EMBEDDING_MODELS
+        )
+        self.known_observations: list[dict] | None = kwargs.pop(
+            "known_observations", None
+        )
 
         self.results: ExperimentResults = ExperimentResults()
         self._exception_handler = ExperimentExceptionHandler(self.event_handler)
@@ -193,12 +206,16 @@ class AI4RAGExperiment:
                         which_doc = doc.metadata.get("document_id") or idx
                         logger.warning("Empty document: %s", which_doc)
                     if not doc.metadata.get("document_id", None):
-                        logger.warning("document_id not provided for document at index: %s", idx)
+                        logger.warning(
+                            "document_id not provided for document at index: %s", idx
+                        )
 
                     proper_docs.append(doc)
 
                 else:
-                    raise ValueError(f"Incorrect type of document provided at index: {idx}.")
+                    raise ValueError(
+                        f"Incorrect type of document provided at index: {idx}."
+                    )
 
         self._documents = proper_docs
 
@@ -271,7 +288,9 @@ class AI4RAGExperiment:
         )
 
         mps = ModelsPreSelector(
-            benchmark_data=self.benchmark_data.get_random_sample(n_records=n_records, random_seed=random_seed),
+            benchmark_data=self.benchmark_data.get_random_sample(
+                n_records=n_records, random_seed=random_seed
+            ),
             documents=self.documents.copy(),
             foundation_models=foundation_models,
             embedding_models=embedding_models,
@@ -280,7 +299,8 @@ class AI4RAGExperiment:
         mps.evaluate_patterns()
 
         selected_models = mps.select_models(
-            n_embedding_models=self.n_mps_embedding_models, n_foundation_models=self.n_mps_foundation_models
+            n_embedding_models=self.n_mps_embedding_models,
+            n_foundation_models=self.n_mps_foundation_models,
         )
 
         logger.info(
@@ -316,7 +336,9 @@ class AI4RAGExperiment:
 
         distance_metric = EmbeddingModels.get_distance_metric(embedding_model.model_id)
         embedding_params_dict = (
-            asdict(embedding_model.params) if is_dataclass(embedding_model.params) else embedding_model.params
+            asdict(embedding_model.params)
+            if is_dataclass(embedding_model.params)
+            else embedding_model.params
         )
         indexing_params = {
             "chunking": chunking_params,
@@ -351,6 +373,9 @@ class AI4RAGExperiment:
                 "user_message_text": user_message_text,
                 "system_message_text": system_message_text,
             },
+            "vector_io_provider_type": self.client.providers.retrieve(
+                self.ogx_vector_io_provider_id
+            ).provider_type,
         }
 
         logger.info("Using retrieval and generation params: %s", rag_params)
@@ -362,9 +387,13 @@ class AI4RAGExperiment:
             return result_score
 
         pattern_name = self._create_pattern_name()
-        logger.info("Using name '%s' for the currently evaluated pattern.", pattern_name)
+        logger.info(
+            "Using name '%s' for the currently evaluated pattern.", pattern_name
+        )
 
-        reuse_collection_name = self._get_reusable_collection_name(indexing_params=indexing_params)
+        reuse_collection_name = self._get_reusable_collection_name(
+            indexing_params=indexing_params
+        )
 
         vector_store = get_vector_store(
             vs_type=self.vector_store_type,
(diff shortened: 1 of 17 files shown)


jakub-walaszczyk:
@filip-komarzyniec please restore changes that introduce formatting for 80 characters per line. 120 has been fine here, and these changes make noise in the PR

ai4rag/rag/foundation_models/ogx.py
```diff
@@ -58,26 +58,42 @@ def params(self, params: dict | OGXModelParameters | None) -> None:
         else:
             self._params = OGXModelParameters()
 
-    def chat(self, messages: list[MessageTyped]) -> list[MessageTyped]:
+    def create_response(self, user_message: str, vector_store_id: str | None = None) -> str:
```
jakub-walaszczyk, line 85: Why do we need to remove `chat()`? Can't we have simply another method added?

ai4rag/core/experiment/experiment.py
```diff
@@ -28,7 +28,11 @@
     get_retrieval_params,
     query_rag,
 )
-from ai4rag.core.hpo.base_optimizer import BaseOptimizer, OptimizationError, OptimizerSettings
+from ai4rag.core.hpo.base_optimizer import (
```
jakub-walaszczyk: Yet another formatting artifact. It does not matter much to add/remove these. Whenever you are trackings changes, these things are just a noise.

ai4rag/core/experiment/experiment.py
```diff
@@ -154,7 +158,12 @@ def __init__(
 
         self.job_id = kwargs.pop("job_id", "ai4rag_job_a0b1c2d3").replace("-", "_")
         self.metrics: Sequence[str] = kwargs.pop(
-            "metrics", (MetricType.ANSWER_CORRECTNESS, MetricType.FAITHFULNESS, MetricType.CONTEXT_CORRECTNESS)
+            "metrics",
+            (
+                MetricType.ANSWER_CORRECTNESS,
+                MetricType.FAITHFULNESS,
+                MetricType.CONTEXT_CORRECTNESS,
+            ),
```
jakub-walaszczyk: I would keep 1 line instead of 6 :/

ai4rag/core/experiment/experiment.py
```diff
@@ -351,6 +361,7 @@ def run_single_evaluation(self, rag_params: RAGParamsType) -> float:
                 "user_message_text": user_message_text,
                 "system_message_text": system_message_text,
             },
+            "vector_io_provider_type": self.client.providers.retrieve(self.ogx_vector_io_provider_id).provider_type,
```
jakub-walaszczyk: So, not each scenario must run with the `client`.
In such case, this will fail for chroma scenarios (please see functional tests).

> Yes chroma is supported within `ai4rag`. And this is kind of nice I think to have this smooth decoupling, so that we can run this with no external vector db.

**Recommendation: **
First instantiate vector store, add this parameter later extracting from the `VectorStore`

OR

Make some logic that will set this based on client's availability

ai4rag/core/experiment/experiment.py
```diff
@@ -381,7 +392,11 @@ def run_single_evaluation(self, rag_params: RAGParamsType) -> float:
             chunk_size = chunking_params.get(AI4RAGParamNames.CHUNK_SIZE)
             chunk_overlap = chunking_params.get(AI4RAGParamNames.CHUNK_OVERLAP)
 
-            chunker = LangChainChunker(method=chunking_method, chunk_size=chunk_size, chunk_overlap=chunk_overlap)
+            chunker = LangChainChunker(
+                method=chunking_method,
+                chunk_size=chunk_size,
+                chunk_overlap=chunk_overlap,
+            )
```
jakub-walaszczyk: 👀

ai4rag/core/experiment/experiment.py
```diff
@@ -525,10 +540,14 @@ def objective_function(space: RAGParamsType) -> float | None:
...
-                name=AI4RAGParamNames.FOUNDATION_MODEL, param_type="C", values=selected_models["foundation_models"]
+                name=AI4RAGParamNames.FOUNDATION_MODEL,
+                param_type="C",
+                values=selected_models["foundation_models"],
             )
             self.search_space[AI4RAGParamNames.EMBEDDING_MODEL] = Parameter(
-                name=AI4RAGParamNames.EMBEDDING_MODEL, param_type="C", values=selected_models["embedding_models"]
+                name=AI4RAGParamNames.EMBEDDING_MODEL,
+                param_type="C",
+                values=selected_models["embedding_models"],
```
jakub-walaszczyk: We really don't need that 😢

> but you can remove `param_type="C"`, it is default anyway

ai4rag/core/experiment/experiment.py
```diff
@@ -624,10 +638,29 @@ def _stream_finished_pattern(
...
+                    "instructions": evaluation_result.rag_params["generation"]["system_message_text"],
+                    "tools": [
+                        {
+                            "type": "file_search",
+                            "vector_store_ids": evaluation_result.collection,
+                        }
+                    ],
+                    "include": ["file_search_call.results"],
+                },
             },
```
jakub-walaszczyk, line 649: Chroma scenario is not considered here. Please make sure that this is properly parsed when `chroma` is used within `ai4rag` context.

ai4rag/core/experiment/experiment.py
```diff
@@ -669,7 +702,9 @@ def _evaluate_response(
         """
 
         logger.info(
-            "Evaluating the RAG Pattern '%s' response using %s.", pattern_name, self.evaluator.__class__.__name__
+            "Evaluating the RAG Pattern '%s' response using %s.",
+            pattern_name,
+            self.evaluator.__class__.__name__,
         )
```
jakub-walaszczyk: 👀

ai4rag/rag/foundation_models/base_model.py
```diff
@@ -72,5 +65,5 @@ def __lt__(self, other: object) -> bool:
         return self.model_id < other.model_id
 
     @abstractmethod
-    def chat(self, messages: list[MessageTyped]) -> list[MessageTyped]:
-        """Chat with the model base on the client capabilities."""
+    def create_response(self, user_message: str, vector_store_id: str | None = None) -> str:
```
jakub-walaszczyk, line 75: Simply add another method please 👀

ai4rag/rag/foundation_models/base_model.py
```diff
@@ -16,13 +16,6 @@
 FoundationModelParamsT = TypeVar("FoundationModelParamsT")
 
 
-class MessageTyped(TypedDict):
-    """Type of the messages used by the client."""
-
-    role: str
-    content: str
```
jakub-walaszczyk, line 23: Does not need to be removed

ai4rag/rag/foundation_models/ogx.py
```diff
@@ -58,26 +58,42 @@ def params(self, params: dict | OGXModelParameters | None) -> None:
...
+            instructions=self.system_message_text,
+            input=user_message,
             max_completion_tokens=self.params.max_completion_tokens,
             temperature=self.params.temperature,
+            tools=tools,
         )
-        response_choices = response_chat.choices
 
-        return response_choices
+        return response.output_text
```
jakub-walaszczyk, line 123: Is simply text expected as output for the responses?
E.g. in chat we expect to keep all the messages in history and that makes it reusable in further scenarios.

ai4rag/rag/vector_store/ogx.py
```diff
@@ -198,7 +201,13 @@ def search(
...
         if include_scores:
             return [
-                (Document(page_content=chunk.content, metadata=chunk.chunk_metadata.to_dict()), score)
+                (
+                    Document(
+                        page_content=chunk.content,
+                        metadata=chunk.chunk_metadata.to_dict(),
+                    ),
+                    score,
+                )
```
jakub-walaszczyk: 👀

dev_utils/mocks.py
```diff
@@ -30,15 +30,24 @@ def __init__(
             context_template_text=context_template_text,
         )
 
-    def chat(self, messages: list[MessageTyped]) -> list[MessageTyped]:
-        # Create a mock response that mimics the structure of real model responses
-        class MockMessage:
-            content = "I cannot answer this question, because I am just a mocked model."
+    def create_response(self, user_message: str, vector_store_id: str | None = None) -> str:
```
jakub-walaszczyk, line 33: New method is recommended instead of overriding `chat`

jakub-walaszczyk (changes requested):
Please follow the requested changes and consider handling chroma within the `ai4rag` context.

ai4rag/rag/foundation_models/ogx.py
filip-komarzyniec: I'll revert that if you feel that  `chat.completions` should still be supported. 

I deleted that as from the KFP POV the `chat.completions` support is to be dropped. With the KFP being the main consumer of `ai4rag` I didn't see the reason for the `chat.completions` to still be supported --- the less code the better.

ai4rag/core/experiment/experiment.py
filip-komarzyniec: All of the formatting changes are introduced by `black` pulled by the `ai4rag` library, these are not my choices and without them the repository checks will not pass. 

Maybe you're comparing those with older black version or have some local settings?

ai4rag/core/experiment/experiment.py
filip-komarzyniec: That's a valid point, thanks. I thought that Chroma requires the `client` instance as well... 

Let me think how to refactor that now.

ai4rag/rag/foundation_models/ogx.py
filip-komarzyniec: The way it works right now in OGX is that `response.output_text` seems to be concatenated str representation of all returned `Item`s in the response. 

In other words, the `output_text` holds a string containing tool calls, CoT reasoning, final answer and any other stuff in the message history. 

I'm running some tests now in order to figure out whether this is a bug on OGX's side or an expected behaviour.

ai4rag/rag/retrieval/retriever.py
```diff
@@ -52,6 +52,11 @@ def __init__(
         self.ranker_k = ranker_k
         self.ranker_alpha = ranker_alpha
 
+    @property
+    def collection_name(self):
+        """Exposes the collection_name this retriever seraches through."""
```
filip-komarzyniec: typo

ai4rag/utils/constants.py
```diff
@@ -112,7 +112,7 @@ class GenerationConstants(metaclass=ConstantMeta):
 class ChatGenerationConstants(metaclass=ConstantMeta):
     """Constants used for setting the generation (inference) parameters for chat models only."""
 
-    MAX_COMPLETION_TOKENS = 2048
+    MAX_TOKENS = 2048
```
jakub-walaszczyk, line 115: If the parameter name is now `max_output_tokens` why do we use `max_tokens`? Wouldn't it be better to reflect the name of the parameter?

filip-komarzyniec: It would,  if we dropped the `chat.completions` support. The problem is that in the older API (completions) this param is called `max_completion_tokens`. In newer API (responses) it's `max_output_tokens`. 

Now that we have some abstract constants class, I've decided to call it just `max_tokens`there because it's consumed by both API interfaces. 
What's more, the `max_tokens` is a deprecated name once used by `chat.completions` API so it's not totally made up by me. 

The whole problem is caused by the fact that we decided earlier to leave the `chat` method support. No one uses it now so maybe it's better to just replace it with responses-aligned methods.

ai4rag/rag/foundation_models/ogx.py
filip-komarzyniec: Most probably this is not an OGX server issue as after changing the model the behaviour also changed to the expected one. 

Why it depends on the model -- I'm not really sure. Maybe we'll have to investigate it further once more reports of such behaviour across different models are logged.

ai4rag/utils/constants.py
jakub-walaszczyk: Let's replace it and keep proper parameters names, Thanks

filip-komarzyniec:
Closing. The requirements have changed and these changes' priority has decreased. Future refactoring PRs might further build on these.

PR #70, opened 2026-06-11
feat(search_space): Introduce minimal required context_length for embedding models (700 tokens)

jakub-walaszczyk:
## Description
Added validation of minimal required context_length for embedding models to make sure experiments will not fail on minial chunk sizes.

## Motivation
Some embedding models support very small context length (e.g. 256 tokens). Such models will keep on failing in ai4rag experiments, as minimal supported chunk size is 512 tokens.

## Changes
- Remove constant artifacts that are no longer valid and used
- Introduce check for embedding model context_length
- Update tests and test new check

## Testing
Via unit tests

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/rag/embedding/ogx.py
+++ b/ai4rag/rag/embedding/ogx.py
@@ -13,6 +13,9 @@ from .base_model import BaseEmbeddingModel
 __all__ = ["OGXEmbeddingModel", "OGXEmbeddingParams"]
 
 
+MIN_CONTEXT_LENGTH = 700
+
+
 @dataclass
 class OGXEmbeddingParams:
     """OGX parameters to be used to create embeddings."""
@@ -40,7 +43,7 @@ class OGXEmbeddingModel(BaseEmbeddingModel[OgxClient, OGXEmbeddingParams]):
 
     @params.setter
     def params(self, params: dict | OGXEmbeddingParams | None) -> None:
-        """Set model params."""
+        """Set model params and validate context length."""
         if params is None:
             self._params = OGXEmbeddingParams()
         elif isinstance(params, OGXEmbeddingParams):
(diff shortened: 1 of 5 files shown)


ai4rag/rag/embedding/ogx.py
```diff
@@ -54,6 +57,12 @@ def params(self, params: dict | OGXEmbeddingParams | None) -> None:
         if self._params.context_length is None:
             self._params.context_length = self._detect_context_length()
 
+        if self._params.context_length < MIN_CONTEXT_LENGTH:
+            raise ValueError(
+                f"Embedding model {self.model_id} supports `context_length` {self._params.context_length}. "
+                f"Minimal required value is {MIN_CONTEXT_LENGTH}. Use embedding model with greater `context_length`."
+            )
+
```
MichalSteczko, line 65: Are we 100% sure that it is always present?
OGX is requiring `embedding_dimension` parameter only. 
Hence I guess that this part should be moved to model validation function.

MichalSteczko (commented):
Please check the comment.

ai4rag/rag/embedding/ogx.py
jakub-walaszczyk: When this parameter is not passed during `OGXEmbeddingModel` instantiation (in `prepare_serach_space` we look for these both parameters in the model's metadata) we make some fallback and try to set both these parameters.

`embedding_dimension` if not provided, it is calculated by sending a single request and checking the length of the generated vector.

`context_length` if not explicitly provided in the metadata is estimated via binary search algorithm. Is is always needed, as this is required for the `chunk_size` configuration per model (please see search space preparation rules).

This rule here is safe. We could place it in different location, but as we check only this one I find this line completely fine. Maybe when we will have more validations, we will move it to another class, logic?

MichalSteczko: Thank you for the explanation.
Separation / decoupling is a great idea here. We can do it in the future.

PR #72, opened 2026-06-15
feat(evaluator): add LLM-as-a-Judge with multi-evaluator dispatch

piotrhm:
## Summary

Adds LLM-as-a-Judge evaluation to the RAG optimization pipeline. A new `LLMaJEvaluator` scores `answer_relevance` using a foundation model as a judge, while `AI4RAGExperiment` is extended to accept multiple evaluators, dispatch metrics by evaluator type, and merge results into a unified score.

## Motivation

Algorithmic metrics (faithfulness, answer_correctness, context_correctness) computed by Unitxt are fast and deterministic but miss nuanced quality signals like helpfulness and on-topic relevance. An LLM judge adds that dimension while Unitxt continues to handle grounding and correctness.

### New modules
- **`llmaj_evaluator.py`** — `LLMaJEvaluator` scores `answer_relevance` via `BaseFoundationModel.chat()` with structured JSON output (1–5 rubric, normalized to [0.0, 1.0]); bootstrap confidence intervals for aggregate scores
- **`judge_selection.py`** — `select_judge_model()` auto-selects the best judge from the generation model pool using a spread-and-stability calibration on a benchmark subset

### Metric additions
- `JUDGE_ANSWER_RELEVANCE` and `OVERALL_SCORE` added to `Metrics`; default optimization metric changed from `faithfulness` to `overall_score`

### Multi-evaluator architecture
- `AI4RAGExperiment` accepts an `evaluators: list[BaseEvaluator]` kwarg (replaces the single `evaluator` parameter)
- Per-pattern evaluation dispatches each metric to its matching evaluator by `EVALUATOR_TYPE`, then merges results before computing derived metrics (e.g. `overall_score`)
- Default metrics are auto-derived from configured evaluators when not explicitly provided
- `BaseEvaluator` gains an `EVALUATOR_TYPE` class attribute and optional `model_id` on `AggregateMetric`

### Optimization pipeline
- `run_rag_optimization` selects a judge model at optimization time and wires both `UnitxtEvaluator` and `LLMaJEvaluator` into the experiment
- `ModelsPreSelector` metric parameter is now optional (defaults to `overall_score`); always evaluates all Unitxt metrics during pre-selection
- Search space preparation simplified — no longer accepts a metric parameter
- Output artifact generation extracted into `_generate_output_artifacts()`

### Removed
- `HybridEvaluator`, `LLMaJConfig`, `score_utils.py` — superseded by the multi-evaluator dispatch
- `openai` optional dependency — LLM judge uses the foundation model abstraction layer
- `llm-as-judge-design.md` design document — superseded by implementation

## Testing

- New unit tests for `LLMaJEvaluator`, judge selection internals (`_spread_and_stability_score`, `_score_judge_candidates`, `calibration_subset_size`), experiment multi-evaluator dispatch, and MPS metric defaulting
- Updated functional and integration tests for the new `evaluators` kwarg and `overall_score` default

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] All checks passing


LukaszCmielowski (dismissed):
> Add mlflow>=3.0.0 and openai>=1.0.0 as optional dependencies (ai4rag[llm-judge])

If there is no need for mlflow server I would not make it optional. LLMaJ should be mandatory setup to make autoRAG work with desired quality.

pyproject.toml
```diff
@@ -54,10 +54,16 @@ Documentation = "https://ibm.github.io/ai4rag/"
 
 
 [project.optional-dependencies]
+llm-judge = [
+    "mlflow>=3.0.0",
+    "openai>=1.0.0",
```
jakub-walaszczyk: It would be best to lock version to enable only patch changes (see ~=x.y.z). OpenAI client is released in the 2.y.z version already, and following this constrain we will enable this major shift as well. Let's use latest compatible version and lock it with `~=`

ai4rag/evaluator/base_evaluator.py
```diff
@@ -77,6 +77,7 @@ class MetricType(metaclass=ConstantMeta):
     ANSWER_CORRECTNESS = "answer_correctness"
     FAITHFULNESS = "faithfulness"
     CONTEXT_CORRECTNESS = "context_correctness"
+    ANSWER_RELEVANCE = "answer_relevance"
```
jakub-walaszczyk: This minimal mention in the docstring to follow used convention would be appreciated

ai4rag/evaluator/__init__.py
```diff
@@ -4,3 +4,11 @@
...
 from ai4rag.evaluator.base_evaluator import BaseEvaluator
 from ai4rag.evaluator.unitxt_evaluator import UnitxtEvaluator
+
+try:
+    from ai4rag.evaluator.mlflow_llm_judge_evaluator import (
+        LLMJudgeConfig,
+        MlflowLLMJudgeEvaluator,
+    )
+except ImportError:
+    pass
```
jakub-walaszczyk: Using silent `pass` when we end up in the import error seems not like an option. We should AT LEAST log something on the `info` level.

ai4rag/evaluator/mlflow_llm_judge_evaluator.py
```diff
@@ -0,0 +1,394 @@
...
+
+    custom_metrics : list[CustomMetricDefinition]
+        Optional list of user-defined custom LLM judge metrics.
+    """
+
+    base_url: str = "https://api.openai.com/v1"
+    api_key: str = ""
+    model: str = "gpt-4o-mini"
+    temperature: float = 0.0
+    custom_metrics: list[CustomMetricDefinition] = field(default_factory=list)
```
jakub-walaszczyk: Please remove this default with open AI URL. We do not want to end upt accidentally sending data to wrong endpoints. APIKEY should be required and base url should be required. same for model, no defaults.

ai4rag/evaluator/mlflow_llm_judge_evaluator.py
```diff
@@ -0,0 +1,394 @@
...
+Respond with ONLY a JSON object (no markdown, no extra text):
+{{"score": <integer 1-5>, "rationale": "<brief explanation>"}}
+
+Where:
+- 1 = completely fails the criterion
+- 2 = mostly fails with some relevant elements
+- 3 = partially meets the criterion
+- 4 = mostly meets with minor gaps
+- 5 = fully meets the criterion
+"""
```
jakub-walaszczyk: For reach comparison and optimization, having 5 points scale seems to be somehow not enough. After normalization we will end up with 0.2 0.4. maybe forcing 10 points scale will make more sense to find right balance. To be discussed?

btw, isn't there a custom mlflow mechanism for that?

jakub-walaszczyk (changes requested):
While the implementation of MlflowEvaluator class is nice there are several things to reconsider:
1. As pipelines-components code is the user of the evalutoar, we should probably keep the logic of metrics selection and validation within `ai4rag`. What I mean is that metrics for llmaaj should have some llmaaj prefix, and evaluator should be instantiated in the Experiment orchestrator class based on the metric name. The question is whether we want to use llmaaj only when explicitly requested, change metrics names for that purpose, use it as default? This should be clarified. 
2. We should definitely remove OpenAI endpoint default. If by any mistake we will send some user data to the undesired URL this will be very inapropriate.
3. There is open question with model used for llmaaj metrics assesment. We assume user will pass model, but the user in the different context is us using it in pipelines-components. How do we select the model? How do we select settings? What is the mapping between user's provided metric and whether llmaaj will or will not be used.

These questions should be partially answered by the top design of how this functionality will be triggered.

Maybe we should calculate all of the metrics (why not) and the llmaaj ones simply start with `llmaaj_` prefix, e.g. (`llmaaj_faithfulness`). In that case we can assume 2 evaluators will be used for the metrics assessment or single combined evaluator.

pyproject.toml
Medokins: locked to: `openai~=1.109.1`

ai4rag/evaluator/__init__.py
Medokins: Completed catch block:
```py
except ImportError as exc:
    logger.info(
        "LLM-as-a-Judge evaluator is unavailable (%s). "
        "Install optional dependencies with: pip install ai4rag[llm-judge]",
        exc,
    )
```

ai4rag/evaluator/mlflow_llm_judge_evaluator.py
Medokins: Removed defaults from this dataclass

jakub-walaszczyk (approved):

PR #74, opened 2026-06-22
Move AutoRAG components from `pipelines-components` code to `ai4rag`

jakub-walaszczyk:
## Description

Extracts the core business logic from the external `pipelines-components` repository into `ai4rag.components`, making the pipeline step functions directly importable, testable, and usable outside of Kubeflow Pipelines. This consolidation introduces four new subpackages — `data`, `optimization`, `assets_generator`, and `utils` — alongside refactors to the experiment result schema and removal of the hardcoded `EmbeddingModels`/`FoundationModels` enums in favor of plain model-ID strings.

## Motivation

The pipeline component business logic (document discovery, text extraction, indexing, search space preparation, RAG optimization, and artefact generation) was previously embedded inside KFP wrapper code in the `pipelines-components` repo. This made it impossible to unit test in isolation, reuse outside of KFP, or evolve alongside the core `ai4rag` library. By moving the logic here, we:

- Enable standalone usage in scripts, notebooks, and tests without KFP dependencies
- Consolidate versioning under the `ai4rag` release lifecycle
- Remove the brittle `FoundationModels`/`EmbeddingModels` enum classes that broke every time a new model was onboarded, replacing them with plain string IDs
- Align the experiment result schema with the latest pattern definition format (field renames: `pattern_name` → `name`, `execution_time` → `duration_seconds`, `datasource_type` → `provider_id`, etc.)

### New packages
- **`ai4rag.components.data`** — `discover_documents()`, `extract_text()`, `index_documents()`, `load_test_data()`
- **`ai4rag.components.optimization`** — `prepare_search_space_report()`, `run_rag_optimization()`, `detect_benchmark_language()`
- **`ai4rag.components.assets_generator`** — `Notebook`, `NotebookCell`, `build_leaderboard_html()`, `build_pattern_json()`, `generate_notebook_from_template()` with bundled `.ipynb` and `.templ` templates
- **`ai4rag.components.utils`** — `create_s3_client()`, `create_ogx_client()`, `load_docling_documents()`
- **`ai4rag.utils.compat`** — `ensure_sqlite3()` centralizing the pysqlite3 patch for RHEL 9 / older sqlite environments

### Refactors
- **Deleted `ai4rag/search_space/src/models.py`** — `FoundationModels` and `EmbeddingModels` enum classes removed; `model_props.py` now uses plain string model IDs directly
- **`OGXEmbeddingParams`** — removed unused fields (`timeout`, `model_type`, `provider_id`, `provider_resource_id`)
- **`AI4RAGExperiment._build_evaluation_payload()`** — updated result schema (field renames, conditional ranker params based on strategy, added `provider_type` lookup, removed `schema_version`/`producer`)

### Dependencies
- `boto3>=1.28` added as core dependency
- `docling-core` bumped from `~=2.74.1` to `~=2.78.1`
- New optional extra: `ai4rag[components]` (adds `multiprocess>=0.70`)
- Version bumped to `0.8.0`

### Documentation
- New user guide: `docs/user-guide/pipeline-components.md`
- New API reference pages for data, optimization, and assets generator modules
- Architecture overview updated with Pipeline Components layer
- Installation guide updated with extras table
- Changelog entry for 0.8.0

## Testing

- **20 new test files** with ~3,400 lines of unit tests covering all new modules:
  - `tests/unit/ai4rag/components/data/` — discovery, extraction, indexing, test data loader
  - `tests/unit/ai4rag/components/optimization/` — language detection, RAG optimization, search space prep
  - `tests/unit/ai4rag/components/utils/` — S3 client, OGX client, docling I/O
  - `tests/unit/ai4rag/assets_generator/` — leaderboard, notebook, pattern builder, templates
  - `tests/unit/ai4rag/utils/test_compat.py` — sqlite3 compatibility patch
- Orchestration modules (`text_extraction.py`, `rag_templates_optimization.py`) are excluded from coverage as they require integration-level setup (S3, multiprocessing, full experiment runs)

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [ ] All checks passing


ai4rag/components/assets_generator/notebook_templates/ogx_indexing_template.ipynb
```diff
@@ -0,0 +1,320 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": "---\n\n## Setup\n\nThis section sets up the notebook environment by installing required packages, importing libraries, and configuring access to S3 storage.\n\n### Install packages\n\nInstall all required Python packages for document processing and RAG operations:\n- **docling**: Document processing and text extraction library\n- **ai4rag**: The AutoRAG framework for building RAG applications"
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": "%pip install docling | tail -n 1\n%pip install 'ai4rag~=0.7.0' | tail -n 1"
```
DorotaDR: Enhancement proposal: Now that the notebooks are in ai4rag, it would be good to parameterize the `ai4rag` version and dynamically fill it with the current release (for example, using `ai4rag.__version__`). This can be added in a separate PR.

jakub-walaszczyk: This is literally what I said in the meeting 50 minutes ago :D

ai4rag/components/data/text_extraction.py
```diff
@@ -0,0 +1,578 @@
...
+        WordFormatOption,
+    )
+
+    ap = _resolve_artifacts_path(None)
+    accel = AcceleratorOptions(device="cpu", num_threads=2)
+
+    pdf_pipeline_options = ThreadedPdfPipelineOptions(
+        artifacts_path=ap,
+        do_ocr=False,
+        do_table_structure=False,
```
DorotaDR: Missing Lukasz's changes from https://github.com/opendatahub-io/pipelines-components/pull/139

(`        do_table_structure=True`)

jakub-walaszczyk: Thanks Dorota! added

ai4rag/components/assets_generator/notebook_templates/ogx_indexing_template.ipynb
jakub-walaszczyk: Added in the ongoing PR. Thanks for motivation!

ai4rag/components/assets_generator/script_templates/create_model_response.py.templ
filip-komarzyniec, line 1: This script is to be deleted as per this jira effort: https://redhat.atlassian.net/browse/RHOAIENG-70297

ai4rag/components/assets_generator/pattern_builder.py
filip-komarzyniec, line 1: Seems unnecessarily complicated to be introducing the whole function just to add `responses_template` and `detected_language` keys. Why these changes cannot be incorporated into `_stream_finished_pattern` function?

ai4rag/components/assets_generator/pattern_builder.py
```diff
@@ -0,0 +1,60 @@
...
+# -----------------------------------------------------------------------------
+def build_pattern_json(
+    pattern: dict,
+    detected_language: dict | None = None,
+) -> dict:
+    """Build a ``pattern.json`` dictionary from evaluation results.
+
+    Normalises parameters from various locations in the evaluation result
+    (indexing params, RAG params, flat attributes) into the canonical flat
+    schema used by downstream consumers (notebook templating, leaderboard).
```
filip-komarzyniec: It does not seem like any normalisaiton but rather enahncing the input pattern based on the information it already contains 🫪 (except for detected_languge str)

ai4rag/components/optimization/rag_templates_optimization.py
```diff
@@ -0,0 +1,386 @@
...
+        Pattern definitions for each evaluated RAG configuration.
+    evaluations : list
+        Raw evaluation result objects from the experiment.
+    """
+
+    patterns: list[dict]
+    evaluations: list
+
+
+def run_rag_optimization(  # pylint: disable=too-many-locals,too-many-arguments,too-many-positional-arguments
```
filip-komarzyniec, line 54: This is the exact body of a `dsl.component`-wrapped function from pipelines-components repository. 
What is the benefit of creating this additional layer of code? To me the simplest and most intuitive approach would be  to:
* move any building blocks to ai4rag (notebooks, utility functions and all such stuff previously defined within `dsl.component`-wrapped function)
* leave the core logic of the component in the pipelines-component repository

With these proposed changes now:
- this function's signature must match exactly the signature of the `dsl.component`-wrapped function. I'm betting it'll diverge quickly
- Any non-ai4rag changes that could have been introduced to the pipelines-component repo directly in the component's function are now tied to the ai4rag's releases and codebase. Seems like putting unnecessary restrictions even for ourselves
-  ..

ai4rag/components/optimization/rag_templates_optimization.py
```diff
@@ -0,0 +1,386 @@
...
+            case {"type_": "generation", **id_to_params}:
+                model_id, params = id_to_params.popitem()
+                return OGXFoundationModel(client=ogx_client, model_id=model_id, params=params)
+            case _:
+                raise ValueError(f"Cannot load the yml-serialized !Model tag: {mapping}")
+
+    yml.add_constructor("!Model", construct_model_instance, Loader=yml.SafeLoader)
+
+
+def _evaluation_result_fallback(eval_data_list: list, evaluation_result: Any) -> list[dict[str, Any]]:
```
filip-komarzyniec, line 251: I believe there are future plans for deleting that? As this logic already lives here in the codebase of ai4rag?

ai4rag/components/utils/docling_io.py
```diff
@@ -0,0 +1,37 @@
+# -----------------------------------------------------------------------------
+# Copyright IBM Corp. 2026
+# SPDX-License-Identifier: Apache-2.0
+# -----------------------------------------------------------------------------
+from pathlib import Path
+
+from docling_core.types.doc.document import DoclingDocument
+
+
+def load_docling_documents(path: str | Path) -> list[DoclingDocument]:
```
filip-komarzyniec, line 10: IT seems like a way too narrowly-scoped function (it loads only .json extensions without the ability to customise it) to create for it a special shared directory, doesn't it?

ai4rag/components/utils/ogx_client.py
filip-komarzyniec, line 1: This is really trash implementation making little sense, e.g.:

- walking the exception chain as if it could have cycles is out of this world 
-  Most of`(ssl.SSLCertVerificationError, httpx.ConnectError, OGXAPIConnectionError)` could not occur in this scenario
- exact substring exception matching is very lousy and prone to quickly fail with future changes
- ...

I once rewrote that part (not sure what happened with my code). Can we at least log it in the backlog to improve this?

ai4rag/utils/compat.py
```diff
@@ -0,0 +1,24 @@
...
+def ensure_sqlite3() -> None:
+    """Patch stdlib ``sqlite3`` with ``pysqlite3-binary`` if available.
+
+    ChromaDB requires sqlite3 >= 3.35. On platforms with an older system
+    sqlite (e.g. RHEL 9), this function swaps the stdlib module with the
+    ``pysqlite3`` wheel so that ChromaDB (and LangChain-Chroma) can work.
+
+    Safe to call multiple times — the patch is idempotent.
+    """
+    if "pysqlite3" in sys.modules or getattr(sys.modules.get("sqlite3"), "__name__", None) == "pysqlite3":
```
filip-komarzyniec: `if "pysqlite3" in sys.modules` is this really correct? 

What this fix has to do is to "trick" `sqlite`-dependent libraries to use `pysqlite` by substituting sys.modules path. 

If both `sqlite` and `pysqlite` are imported, the `if "pysqlite3" in sys.modules` is true but `sqlite` will still point to `sqlite`... resulting in crash.

pyproject.toml
```diff
@@ -58,10 +61,10 @@ dev = [
     "ai4rag[test]",
     "ai4rag[code_check]",
     "ai4rag[docs]",
+    "ai4rag[components]",
```
filip-komarzyniec: How does it work exactly? I do not see anywhere a `components` extras?

filip-komarzyniec (commented):
a few reflections from my side

ai4rag/components/assets_generator/script_templates/create_model_response.py.templ
jakub-walaszczyk: I will do it within these 2 PRs. Seems related. Thank you

ai4rag/components/assets_generator/pattern_builder.py
jakub-walaszczyk: You are right, I wanted to make it in 2 steps:
1. Switch and simplify
2. Further enhancements.

Now language information, due to incorrect implementation is not propagated to the experiment. To add it in `_stream_finished_pattern` we first need to fix multilingual implementation. With that fix, I will remove this `pattern_builder.py` and keep this logic in the Experiment class.

ai4rag/components/optimization/rag_templates_optimization.py
jakub-walaszczyk: I did it with all components. If you introduce some helper functions the problem `this function's signature must match exactly the signature of the dsl.component-wrapped function. I'm betting it'll diverge quickly` we will have the same problem, as these utils functions will strongly base on the input.

Additionally this is not 1:1. this is around 40% of the code previously contained within the component. It got simplified and now we have some easier testing here. We can break it into pieces with next releases and if needed, switch to blocks instead of this single blob. But this requires much more effort to find the right balance.

Have you got some proposition? Let's open an issue or discuss it. And for me the better option would be to keep it in separate PR.

ai4rag/components/optimization/rag_templates_optimization.py
jakub-walaszczyk: Yes, the goal is to simplify internal ai4rag codebase in the next steps as it does not affect public API.
The longer I work on this PR, the longer PR in pipelines-components is open, the harder the merges are.

I would like to add any other enhancements in the next steps.

ai4rag/components/utils/docling_io.py
jakub-walaszczyk: It has some checks and cases. It does not fit other utils modules and is heavily used in other functions. What is your proposal?

ai4rag/components/utils/ogx_client.py
jakub-walaszczyk: Please open an issue. This seems to be 1:1 implementation of what it was in the pipelines. Open an issue, assign it to yourself or myself, and once this PR is merged, open another one targeting this please.

I agree this implementation is trashy, but the more changes I make here the greater is the chance for regression.

pyproject.toml
jakub-walaszczyk: Dev artifact. Great catch, thanks

ai4rag/utils/compat.py
jakub-walaszczyk: Thank you Filip. Fix was added.

ai4rag/components/optimization/rag_templates_optimization.py
filip-komarzyniec: of course, this is non-blocking matter so we can continue the discussion elsewhere. But it will be good to have that discussion as those decisions will significantly affect the ease of future development and maintenance. 
I'll open some jira effort for that

ai4rag/components/utils/ogx_client.py
filip-komarzyniec: I'll try to dig up my previous changes to this part of code and open them up once again for review (really, what happened to them? 🤔)

ai4rag/components/utils/docling_io.py
filip-komarzyniec: just a quick observation. Taking into account the number of times this function is called throughout different files, it of course makes sense to have it here. 🙌🏼
On the other hand, the docstrings take more lines than the code 🫪

Let's not dig into it, if docling_io module will grow in the future than I'm perfectly fine. If not, maybe we'll see a good opportunity to remove it and incorporate those couple lines where they are actually used.

filip-komarzyniec (approved):
/lgtm 🤘🏼 crazy piece of good work 👷🏼‍♂️

PR #80, opened 2026-06-26
Increased customizability of the component functions

### LINKED ISSUE #79: [feat]: Expose configurable max_threads for benchmark RAG queries in AI4RAGExperiment
## Summary

Make the concurrency limit for parallel benchmark inference (`query_rag` → `ThreadPoolExecutor`) configurable at the **experiment** and **`run_rag_optimization`** entry points, so OpenShift AI AutoRAG presets can tune generation load (e.g. `speed` = 10, `balanced` = 4) without forking ai4rag.

---

## Background

During each pattern evaluation, ai4rag runs all benchmark questions concurrently via [`query_rag()`](https://github.com/IBM/ai4rag/blob/main/ai4rag/core/experiment/utils.py):

- `ThreadPoolExecutor(max_workers=max_threads)` — **default `max_threads=10`**
- Each worker: retrieve → build prompt from up to `number_of_chunks` chunk strings → **one** `foundation_model.chat()` call

Call sites today **do not pass `max_threads`**:

- [`AI4RAGExperiment._evaluate_pattern()`](https://github.com/IBM/ai4rag/blob/main/ai4rag/core/experiment/experiment.py) → `query_rag(rag=..., questions=...)`
- [`ModelsPreSelector`](https://github.com/IBM/ai4rag/blob/main/ai4rag/core/experiment/mps.py) → same
mentation/components/autorag/features/experiment_settings.md#balanced) (`speed`: 10, `balanced`: 4).

**Related (separate issue):** pipelines-components maps `preset` → ai4rag `optimization_settings` once this lands.

---

## Acceptance criteria

- [ ] `AI4RAGExperiment(..., max_threads=N)` limits concurrent benchmark queries to `N`.
- [ ] `run_rag_optimization(..., optimization_settings={"max_threads": N})` applies the same limit.
- [ ] `ModelsPreSelector` respects the experiment’s `max_threads` during pre-selection.
- [ ] Default remains **10** when unset (no behavior change for existing callers).
- [ ] Invalid values (`0`, negative, non-int) raise clear `ValueError` / `TypeError`.
- [ ] Unit tests for experiment + `run_rag_optimization` settings plumbing (mock `query_rag` and assert `max_threads` passed).
- [ ] Docs updated.

---

## Implementation notes

**Files (expected touch set):**

| File | Change |
|------|--------|
| `ai4rag/core/experiment/experiment.py` | Store `max_threads`; pass to `query_rag` |
| `ai4rag/core/experiment/mps.py` | Pass `max_threads` to `query_rag` |
| `ai4rag/components/optimization/rag_templates_optimization.py` | Parse `max_threads` from `optimization_settings` |
| `tests/unit/ai4rag/components/optimization/test_rag_optimization.py` | Settings validation + forwarding |
| `tests/unit/ai4rag/core/experiment/test_experiment.py` (or new) | Experiment forwards `max_threads` |

---

## Test plan

1. `query_rag` with `max_threads=2` and 5 questions → at most 2 concurrent `generate()` calls (existing / extend).
2. `AI4RAGExperiment` with `max_threads=3` → mock `query_rag`, verify call arg.
3. `run_rag_optimization(optimization_settings={"max_threads": 4})` → experiment created with `max_threads=4`.
4. Omit `max_threads` → behavior unchanged (10).
5. `max_threads=0` → `ValueError`.

---

## Follow-up (pipelines-components)

After ai4rag release:

- Map `preset=speed` → `optimization_settings["max_threads"] = 10`
- Map `preset=balanced` → `optimization_settings["max_threads"] = 4`
- Log `max_threads` on MLflow parent run params (optional)

### Steps to Reproduce

Run ai4rag experiment.

### 1. `AI4RAGExperiment`

Add optional constructor kwarg:

```python
max_threads: int = 10
```

- Validate: `max_threads >= 1` (optionally cap upper bound, e.g. `<= 32`, to avoid accidental overload).
- Pass through to every `query_rag(..., max_threads=self.max_threads)` in the experiment path.

### 2. `ModelsPreSelector`

Accept the same setting (constructor arg or shared experiment config) so model pre-selection uses the same concurrency as the main optimization loop.

### 3. `run_rag_optimization()`

Extend `optimization_settings` dict:

```python
optimization_settings = {
    "metric": "faithfulness",
    "max_number_of_rag_patterns": 8,
    "max_threads": 10,  # new, optional
}
```

Validate type/range in `_validate_optimization_settings()` and forward to `AI4RAGExperiment` / `ModelsPreSelector` as appropriate.

### 4. Documentation

- Update `docs/architecture/data-flow.md` and/or `docs/user-guide/pipeline-components.md`: experiment-level `max_threads`, defaults, preset-oriented guidance.
- Mention in changelog.

**Naming:** keep **`max_threads`** (matches `query_rag` and product docs). Avoid introducing a second name unless aliasing is required for backward compatibility.

### Actual Behavior

---

## Current behavior

| Layer | `max_threads` |
|-------|----------------|
| `query_rag()` | Parameter exists, default **10** |
| `AI4RAGExperiment` | Not exposed; always default |
| `run_rag_optimization()` | Not exposed |
| `ModelsPreSelector` | Not exposed |

Unit tests already cover `query_rag(..., max_threads=2)` in `tests/unit/ai4rag/core/experiment/test_utils.py`.

### Environment

```markdown
- ai4rag version:
- Python version:
- OGX version:
- Vector Store (Milvus/ChromaDB):
- Operating System:
```

### Error Logs

```shell

```

### Additional Context

_No response_

filip-komarzyniec:
## Description
Aligned component functions to recent changes in customer-facing AutoRAG pipelines signature: https://github.com/opendatahub-io/pipelines-components/pull/153
The recent changes mainly add `preset` attribute which should accordingly affect the internal optimization engine.
To what extent, it is described further here: https://github.com/LukaszCmielowski/architecture-decision-records/blob/autox_docs_updates/documentation/components/autorag/features/experiment_settings.md

## Motivation
- Give customers greater control over resources used
- allow both quick try-outs on low resources and longer, way more detailed runs on stronger hardware

## Changes
- multiple functions' signatures changed according to the linked ADR document

## Testing
yes

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] All checks passing

### DIFF
--- a/ai4rag/components/data/text_extraction.py
+++ b/ai4rag/components/data/text_extraction.py
@@ -68,6 +68,7 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
     error_tolerance: float | None = None,
     max_extraction_workers: int | None = None,
     docling_artifacts_path: str | None = None,
+    do_table_structure: bool = False,
 ) -> ExtractionResult:
     """Download documents from S3 and extract text using Docling.
 
@@ -104,6 +105,11 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
     docling_artifacts_path
         Path to pre-downloaded Docling model artifacts for offline use.
         Falls back to ``DOCLING_ARTIFACTS_PATH`` environment variable.
+    do_table_structure
+        Whether Docling should run TableFormer to reconstruct table
+        rows and columns from detected PDF layout.  ``False`` (default)
+        disables table structure parsing; ``True`` enables it to achieve
+        increased accuracy of extracted data.
 
     Returns
     -------
@@ -129,6 +135,8 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
     s3_creds = _resolve_s3_credentials(s3_endpoint, s3_access_key, s3_secret_key, s3_region)
     artifacts_path = _resolve_artifacts_path(docling_artifacts_path)
 
+    _logger.info("Docling table structure parsing: %s", do_table_structure)
+
     documents = sorted(documents, key=lambda d: d.get("size_bytes", 0), reverse=True)
 
     effective_workers = _effective_worker_count(max_extraction_workers)
@@ -148,6 +156,7 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
         mp_context.Pool(
             processes=effective_workers,
             initializer=_text_extraction_pool_initializer,
+            initargs=(do_table_structure,),
         ) as process_pool,
     ):
         download_start = time.perf_counter()
@@ -335,7 +344,7 @@ def _resolve_artifacts_path(explicit: str | None) -> Path | None:
     return p
 
 
-def _build_docling_format_options() -> dict:
+def _build_docling_format_options(do_table_structure: bool = False) -> dict:
     """Build Docling pipeline format options for each supported input format."""
     ap = _resolve_artifacts_path(None)
     accel = AcceleratorOptions(device="cpu", num_threads=2)
@@ -343,7 +352,7 @@ def _build_docling_format_options() -> dict:
     pdf_pipeline_options = ThreadedPdfPipelineOptions(
         artifacts_path=ap,
         do_ocr=False,
-        do_table_structure=True,
+        do_table_structure=do_table_structure,
         accelerator_options=accel,
     )
     paginated_pipeline_options = PaginatedPipelineOptions(
@@ -361,7 +370,7 @@ def _build_docling_format_options() -> dict:
     }
 
 
-def _text_extraction_pool_initializer() -> None:
+def _text_extraction_pool_initializer(do_table_structure: bool = False) -> None:
     """Pool initializer that creates a ``DocumentConverter`` per worker process."""
     os.environ["TQDM_DISABLE"] = "1"
     os.environ.setdefault("OPENBLAS_NUM_THREADS", "1")
@@ -383,7 +392,7 @@ def _text_extraction_pool_initializer() -> None:
 
     mod = sys.modules[__name__]
     # pylint: disable=protected-access
-    mod._mp_worker_converter = DocumentConverter(format_options=_build_docling_format_options())
+    mod._mp_worker_converter = DocumentConverter(format_options=_build_docling_format_options(do_table_structure))
     worker_log.debug(
         "Worker pid=%s: DocumentConverter ready (%.1fs)",
         worker_pid,
--- a/ai4rag/components/optimization/rag_templates_optimization.py
+++ b/ai4rag/components/optimization/rag_templates_optimization.py
@@ -60,6 +60,7 @@ def run_rag_optimization(  # pylint: disable=too-many-locals,too-many-arguments,
     test_data_key: str = "",
     input_data_key: str = "",
     optimization_settings: dict | None = None,
+    max_threads: int = 10,
 ) -> OptimizationResult:
     """Run a full AI4RAG optimization experiment and generate output artefacts.
 
@@ -90,6 +91,11 @@ def run_rag_optimization(  # pylint: disable=too-many-locals,too-many-arguments,
     optimization_settings
         Optional dictionary with ``"metric"`` and/or
         ``"max_number_of_rag_patterns"`` overrides.
+    max_threads
+        Maximum number of concurrent threads used when querying the
+        RAG service during benchmark evaluation.  Lower values reduce
+        per-request concurrency (useful when each request carries more
+        retrieved context).  Defaults to ``10``.
 
     Returns
     -------
@@ -155,6 +161,7 @@ def run_rag_optimization(  # pylint: disable=too-many-locals,too-many-arguments,
         documents=documents,
         optimization_metric=optimization_metric,
         ogx_vector_io_provider_id=vector_io_provider_id,
+        max_threads=max_threads,
     )
 
     # --- Run the optimization loop ---
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -112,6 +112,9 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
     top_k_embedding: int = _DEFAULT_TOP_K_EMBEDDING,
     sample_size: int = _DEFAULT_SAMPLE_SIZE,
     random_seed: int = _DEFAULT_SEED,
+    chunking_methods: list[str] | None = None,
+    contextual_enrichment_enabled: bool | None = None,
+    max_threads: int = 10,
 ) -> SearchSpaceReport:
     """Run model pre-selection and prepare a search-space report.
 
@@ -146,6 +149,20 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         Number of benchmark records sampled for model pre-selection.
     random_seed
         Seed for reproducible sampling.
+    chunking_methods
+        When provided, constrains the ``chunking_method`` dimension of the
+        search space to only these methods (e.g. ``["recursive"]`` or
+        ``["hybrid"]``).  ``None`` uses the platform defaults (both
+        ``"recursive"`` and ``"hybrid"``).
+    contextual_enrichment_enabled
+        Whether LLM contextual enrichment should be applied at index
+        time.  Stored as metadata in the report for downstream
+        consumption.  ``None`` omits it from the report.
+    max_threads
+        Maximum number of concurrent threads used when querying the
+        RAG service during benchmark evaluation.  Lower values reduce
+        per-request concurrency (useful when each request carries more
+        retrieved context).  Defaults to ``10``.
 
     Returns
     -------
@@ -165,6 +182,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
 
     _validate_model_list(embedding_models, "embedding_models")
     _validate_model_list(generation_models, "generation_models")
+    _validate_chunking_methods(chunking_methods)
 
     # Build payload and create search space via OGX
     payload: dict[str, list[dict[str, str]]] = {}
@@ -191,6 +209,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
             foundation_models=search_space._search_space["foundation_model"].values,  # pylint: disable=protected-access
             embedding_models=search_space._search_space["embedding_model"].values,  # pylint: disable=protected-access
             metric=metric,
+            max_threads=max_threads,
         )
         mps.evaluate_patterns()
         selected = mps.select_models(
@@ -215,6 +234,14 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
     }
     verbose_repr["foundation_model"] = [_serialize_model(m) for m in selected_models["foundation_model"]]
     verbose_repr["embedding_model"] = [_serialize_model(m) for m in selected_models["embedding_model"]]
+
+    if chunking_methods is not None:
+        allowed = set(chunking_methods)
+        verbose_repr["chunking_method"] = [m for m in verbose_repr["chunking_method"] if m in allowed]
+        _logger.info("Chunking methods constrained to: %s", verbose_repr["chunking_method"])
+
+    if contextual_enrichment_enabled is not None:
+        _logger.info("Contextual enrichment enabled: %s", contextual_enrichment_enabled)
 
     return SearchSpaceReport(
         search_space=verbose_repr,
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -255,3 +255,16 @@ def _validate_model_list(models: list[str] | None, name: str) -> None:
     for i, m in enumerate(models):
         if not m:
             raise TypeError(f"{name}[{i}] must be a non-empty string.")
+
+
+def _validate_chunking_methods(methods: list[str] | None) -> None:
+    """Validate that chunking methods, if provided, are non-empty strings."""
+    if methods is None:
+        return
+    if not isinstance(methods, list):
+        raise TypeError("chunking_methods must be a list.")
+    if not methods:
+        raise ValueError("chunking_methods must not be empty when provided.")
+    for i, m in enumerate(methods):
+        if not isinstance(m, str) or not m.strip():
+            raise TypeError(f"chunking_methods[{i}] must be a non-empty string.")
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -113,8 +113,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
     sample_size: int = _DEFAULT_SAMPLE_SIZE,
     random_seed: int = _DEFAULT_SEED,
     chunking_methods: list[str] | None = None,
-    contextual_enrichment_enabled: bool | None = None,
-    max_threads: int = 10,
+    inference_max_threads: int = 10,
 ) -> SearchSpaceReport:
     """Run model pre-selection and prepare a search-space report.
 
@@ -154,11 +153,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         search space to only these methods (e.g. ``["recursive"]`` or
         ``["hybrid"]``).  ``None`` uses the platform defaults (both
         ``"recursive"`` and ``"hybrid"``).
-    contextual_enrichment_enabled
-        Whether LLM contextual enrichment should be applied at index
-        time.  Stored as metadata in the report for downstream
-        consumption.  ``None`` omits it from the report.
-    max_threads
+    inference_max_threads
         Maximum number of concurrent threads used when querying the
         RAG service during benchmark evaluation.  Lower values reduce
         per-request concurrency (useful when each request carries more
@@ -209,7 +204,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
             foundation_models=search_space._search_space["foundation_model"].values,  # pylint: disable=protected-access
             embedding_models=search_space._search_space["embedding_model"].values,  # pylint: disable=protected-access
             metric=metric,
-            max_threads=max_threads,
+            max_threads=inference_max_threads,
         )
         mps.evaluate_patterns()
         selected = mps.select_models(
@@ -240,14 +235,10 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         unsupported = [m for m in chunking_methods if m not in available]
         if unsupported:
             raise ValueError(
-                f"Unsupported chunking methods: {unsupported!r}. "
-                f"Available methods: {sorted(available)!r}."
+                f"Unsupported chunking methods: {unsupported!r}. " f"Available methods: {sorted(available)!r}."
             )
         verbose_repr["chunking_method"] = chunking_methods
         _logger.info("Chunking methods constrained to: %s", verbose_repr["chunking_method"])
-
-    if contextual_enrichment_enabled is not None:
-        _logger.info("Contextual enrichment enabled: %s", contextual_enrichment_enabled)
 
     return SearchSpaceReport(
         search_space=verbose_repr,
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -236,8 +236,14 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
     verbose_repr["embedding_model"] = [_serialize_model(m) for m in selected_models["embedding_model"]]
 
     if chunking_methods is not None:
-        allowed = set(chunking_methods)
-        verbose_repr["chunking_method"] = [m for m in verbose_repr["chunking_method"] if m in allowed]
+        available = set(verbose_repr["chunking_method"])
+        unsupported = [m for m in chunking_methods if m not in available]
+        if unsupported:
+            raise ValueError(
+                f"Unsupported chunking methods: {unsupported!r}. "
+                f"Available methods: {sorted(available)!r}."
+            )
+        verbose_repr["chunking_method"] = chunking_methods
         _logger.info("Chunking methods constrained to: %s", verbose_repr["chunking_method"])
 
     if contextual_enrichment_enabled is not None:
--- a/ai4rag/core/experiment/experiment.py
+++ b/ai4rag/core/experiment/experiment.py
@@ -161,6 +161,7 @@ class AI4RAGExperiment:
         )
         self.n_mps_embedding_models = kwargs.pop("n_mps_embedding_models", ModelsPreSelector.DEFAULT_N_EMBEDDING_MODELS)
         self.known_observations: list[dict] | None = kwargs.pop("known_observations", None)
+        self.max_threads: int = kwargs.pop("max_threads", 10)
 
         self.results: ExperimentResults = ExperimentResults()
         self._exception_handler = ExperimentExceptionHandler(self.event_handler)
@@ -442,8 +443,7 @@ class AI4RAGExperiment:
         )
 
         inference_response = query_rag(
-            rag=rag_pattern,
-            questions=list(self.benchmark_data.questions),
+            rag=rag_pattern, questions=list(self.benchmark_data.questions), max_threads=self.max_threads
         )
         result_scores, evaluation_data = self._evaluate_response(
             inference_response=inference_response,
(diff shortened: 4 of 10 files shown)


ai4rag/core/experiment/utils.py
```diff
@@ -74,7 +74,7 @@ class RAGRetrievalParamsType(TypedDict):
     ranker_alpha: int | float
 
 
-def query_rag(rag: BaseRAGTemplate, questions: list[str], max_threads: int = 10) -> list[dict[str, Any]]:
+def query_rag(rag: BaseRAGTemplate, questions: list[str], max_threads) -> list[dict[str, Any]]:
```
Mateusz-Switala: Looks like breaking change... Do we want to do this?

ai4rag/components/data/text_extraction.py
```diff
@@ -335,15 +344,15 @@ def _resolve_artifacts_path(explicit: str | None) -> Path | None:
...
+def _build_docling_format_options(do_table_structure: bool = False) -> dict:
     """Build Docling pipeline format options for each supported input format."""
     ap = _resolve_artifacts_path(None)
     accel = AcceleratorOptions(device="cpu", num_threads=2)
 
     pdf_pipeline_options = ThreadedPdfPipelineOptions(
         artifacts_path=ap,
         do_ocr=False,
-        do_table_structure=True,
+        do_table_structure=do_table_structure,
```
Mateusz-Switala, line 355: Looks like breaking change: before the default value was True, now it is False. However, now the logic looks better. If we are good with such chanages then ok.

Mateusz-Switala (commented):
# Claude Code Review: `RHOAIENG-71884` — Preset Parameter Support for Auto-RAG Pipeline

**Branch:** `RHOAIENG-71884-implement-preset-parameter-support-for-auto-rag-pipeline`
**Base:** `main`
**Commit:** `61c486c` — "Increased customizability of the component functions + related tests tweaks"

---

## 1. High-Level Summary

**Goal:** Expose three previously hardcoded or absent parameters as configurable presets across the Auto-RAG pipeline:

1. `do_table_structure` — lets callers opt-in to Docling's TableFormer PDF table parsing (previously always on, now exposed)
2. `max_threads` — lets callers cap `ThreadPoolExecutor` concurrency during RAG inference; previously a local default inside `query_rag`, now threaded up through 4 call layers
3. `chunking_methods` / `contextual_enrichment_enabled` — new search-space filtering and metadata fields in the preparation step

**Scope:** 10 files touched, 275 lines added / 17 deleted. The pattern is clean additive plumbing, but three issues carry real correctness risk and two create structural fragility.

---

### `[HIGH]` — `do_table_structure` default silently regressed from `True` to `False` — CONFIRMED

**File:** `ai4rag/components/data/text_extraction.py:355`

Before this PR, `_build_docling_format_options` always built `ThreadedPdfPipelineOptions(do_table_structure=True)` — table structure (TableFormer) was unconditionally on. The diff parameterises the flag but gives it a default of `False`:

```python
# Before (hardcoded on):
ThreadedPdfPipelineOptions(..., do_table_structure=True)

# After (off by default):
def _build_docling_format_options(do_table_structure: bool = False) -> dict:
    ...
    ThreadedPdfPipelineOptions(..., do_table_structure=do_table_structure)
```

Every existing caller of `extract_text()` that omits `do_table_structure` — including the generated `ogx_indexing_template.ipynb` notebook — silently loses table parsing with no warning, error, or migration notice. Extraction quality for PDF tables degrades immediately in production.

**Fix:** Flip the default to `True` to preserve the existing behavior. If the intent is to change the default policy (table parsing is expensive), that should be a separate explicit change with a changelog entry and callers updated or warned.

---

### `[HIGH]` — `query_rag` removes `max_threads` default, breaking the public API — PLAUSIBLE

**File:** `ai4rag/core/experiment/utils.py:77`

```python
# Before — two-argument call works:
def query_rag(rag: BaseRAGTemplate, questions: list[str], max_threads: int = 10) -> list[dict[str, Any]]:

# After — max_threads is now a required positional argument with no type:
def query_rag(rag: BaseRAGTemplate, questions: list[str], max_threads) -> list[dict[str, Any]]:
```

`query_rag` is listed in `__all__` and is part of the public API. Any external code calling `query_rag(rag, questions)` now gets:

```
TypeError: query_rag() missing 1 required positional argument: 'max_threads'
```

All internal callers (`mps.py`, `experiment.py`) were updated to pass `max_threads=self.max_threads` explicitly, so the repo itself compiles and tests pass — but external consumers break at runtime with no deprecation warning. The type annotation `int` was also dropped, making `max_threads` implicitly `Any`, which removes static analysis coverage and enables the silent hazard below.

**Fix:** Restore the default — `max_threads: int = 10` — and restore the type annotation. The four call layers that now explicitly pass `self.max_threads` can continue to do so; the default acts as a safety net for external callers and documents the canonical value.

---

### `[HIGH]` — Empty `chunking_method` list after caller-supplied filter, no post-filter validation — PLAUSIBLE

**File:** `ai4rag/components/optimization/search_space_preparation.py:272`

```python
if chunking_methods is not None:
    allowed = set(chunking_methods)
    verbose_repr["chunking_method"] = [m for m in verbose_repr["chunking_method"] if m in allowed]
    _logger.info("Chunking methods constrained to: %s", verbose_repr["chunking_method"])
```

If `chunking_methods=["recursive"]` but OGX's search space only contains `["hybrid"]`, the list comprehension produces `[]`. That empty list is serialised to YAML and subsequently reconstructed as a `Parameter("chunking_method", "C", values=[])` — a search-space dimension with zero candidates, causing the optimiser to either crash or silently produce a degenerate run.

`_validate_chunking_methods` only validates the caller's input list; it does not validate the post-filter result. There are no tests for this scenario.

**Fix:** After the filter, assert that the result is non-empty:

```python
filtered = [m for m in verbose_repr["chunking_method"] if m in allowed]
if not filtered:
    available = verbose_repr.get("chunking_method", [])
    raise ValueError(
        f"chunking_methods={chunking_methods!r} produced an empty intersection with "
        f"available methods {available!r}."
    )
verbose_repr["chunking_method"] = filtered
```

---

### `[MEDIUM]` — `max_threads=None` reaches `ThreadPoolExecutor` with no guard — PLAUSIBLE

**File:** `ai4rag/core/experiment/utils.py:107`

Removing the `int` type annotation means static type checkers can no longer flag `None` being passed. If any future caller passes `max_threads=None`, `ThreadPoolExecutor(max_workers=None)` uses `min(32, os.cpu_count() + 4)` threads — potentially opening 36+ connections to the RAG inference endpoint instead of 10, causing rate-limiting or connection exhaustion. No runtime guard is present inside `query_rag`.

Current callers are safe (`kwargs.pop("max_threads", 10)` always yields an `int`), but the guard was removed rather than moved.

**Fix:** Restore the `int` type annotation. Optionally add an assertion: `assert isinstance(max_threads, int) and max_threads >= 1`.

---

### `[MEDIUM]` — Fragile "pop known metadata keys" pattern — PLAUSIBLE

**File:** `ai4rag/components/optimization/rag_templates_optimization.py:140–141`

```python
detected_language: dict[str, str] | None = search_space_raw.pop("detected_language", None)
search_space_raw.pop("contextual_enrichment_enabled", None)  # NEW

search_space = AI4RAGSearchSpace(
    params=[Parameter(param, "C", values=values) for param, values in search_space_raw.items()]
)
```

`SearchSpaceReport.save_yaml()` writes metadata fields flat into the same YAML dict as search-space parameters. The consumer must maintain a parallel allowlist of metadata keys to `.pop()` before iterating the rest as `Parameter` objects. Adding `contextual_enrichment_enabled` to `SearchSpaceReport` required a matching `.pop()` here — done this time, but there is no compile-time or test-time enforcement.

A third metadata field added to `SearchSpaceReport` without a corresponding `.pop()` silently becomes a `Parameter("new_field", "C", values=<bool>)`, causing wrong or crashed optimization.

**Fix:** Give the YAML a structural separation — write parameters under a `params:` sub-key and metadata at the top level. The consumer then reads `report["params"]` directly and new metadata fields are structurally isolated. Alternatively, define a sentinel set as the single source of truth:

```python
_METADATA_KEYS: frozenset[str] = frozenset({"detected_language", "contextual_enrichment_enabled"})
for key in _METADATA_KEYS:
    search_space_raw.pop(key, None)
```

---

### `[MEDIUM]` — Magic value `10` for `max_threads` duplicated in four places

**Files:** `mps.py:130`, `experiment.py:164`, `search_space_preparation.py:146`, `rag_templates_optimization.py:64`

The prior design encoded the default once in `query_rag`'s signature. This PR strips it and replicates it in all four callers. If the default needs to change, it must be updated in four files in sync; mismatching them means MPS pre-selection and main optimization run at different concurrency levels, making model comparisons invalid.

**Fix:** Define a module-level constant — e.g., `_DEFAULT_MAX_THREADS: int = 10` in `utils.py` — and reference it from all four call sites, or restore the default to `query_rag`'s signature (preferred).

---

### `[LOW]` — `_validate_chunking_methods` duplicates `_validate_model_list` with minor semantic variations

**File:** `ai4rag/components/optimization/search_space_preparation.py:286,297`

The two validators share identical structure but diverge subtly: `_validate_model_list` uses `if not m:` (falsy check), while `_validate_chunking_methods` uses `not isinstance(m, str) or not m.strip()` (stricter — catches whitespace-only strings). `_validate_chunking_methods` also adds an empty-list guard that `_validate_model_list` lacks.

These differences are real but small. A single parameterised helper (`_validate_string_list(items, name, allow_empty=True, require_strip=False)`) would unify them and prevent future drift.

---

### `[LOW]` — `ModelsPreSelector.__init__` mixes `kwargs.get` and `kwargs.pop` — inconsistent ownership

**File:** `ai4rag/core/experiment/mps.py:117,130`

```python
self.evaluator = kwargs.get("evaluator", UnitxtEvaluator())   # not consumed
...
self.max_threads = kwargs.pop("max_threads", 10)               # consumed
```

`evaluator` remains in `kwargs` after `__init__`; `max_threads` is removed. If `kwargs` is ever forwarded (e.g., `super().__init__(**kwargs)`), the stale `evaluator` key causes an unexpected-keyword-argument error. Unlike `AI4RAGExperiment`, which logs remaining unknown kwargs at line 169–170, `ModelsPreSelector` has no such guard.

**Fix:** Change `kwargs.get("evaluator", ...)` to `kwargs.pop("evaluator", ...)` for consistency.

---

### `[LOW]` — Post-hoc `chunking_method` filter applied after the OGX round-trip

**File:** `ai4rag/components/optimization/search_space_preparation.py:270–273`

`prepare_search_space_with_ogx()` is called unconditionally with the full search space, then `chunking_method` values are filtered from the local result. The OGX network call pays the full cost regardless of the constraint. The structurally correct fix is to pass the constraint into `prepare_search_space_with_ogx` so the dimension is bounded before the request is made, not after.

---

## 4. Test Coverage Gaps

### `[MEDIUM]` — `test_max_threads_is_accepted` does not test what its name claims — CONFIRMED

**File:** `tests/unit/ai4rag/components/optimization/test_rag_optimization.py:355`

```python
def test_max_threads_is_accepted(self, mock_ogx_client):
    with pytest.raises(ValueError, match="non-empty string"):
        run_rag_optimization(
            ..., vector_io_provider_id="",   # fires before max_threads is touched
            max_threads=4,
        )
```

The `ValueError` fires from `vector_io_provider_id=""` validation before `max_threads` is ever read. If `max_threads` were silently renamed while the empty-string validation remains, this test still passes and the regression goes undetected.

**Fix:** Either use a valid `vector_io_provider_id` and let execution reach the threading machinery (mocking `AI4RAGExperiment`), or rename the test to `test_max_threads_kwarg_does_not_shadow_validation_error`.

---

### `[MEDIUM]` — No test for empty-intersection `chunking_methods` filter

There is no test that passes `chunking_methods=["recursive"]` when the search space only contains `["hybrid"]`. This is the most dangerous edge case of the new feature and has zero test coverage at any level.

---

### `[LOW]` — `test_max_threads_has_default_of_ten` tests a signature detail, not observable behavior

**File:** `tests/unit/ai4rag/components/optimization/test_rag_optimization.py:347`

Using `inspect.signature` to assert `param.default == 10` couples the test to the implementation detail that the default lives in the function signature. If the default were moved to a dataclass or config object, this test breaks even though the observable behavior is unchanged.

---

## 5. Actionable Fixes Summary

| # | Severity | File | Action |
|---|----------|------|--------|
| 1 | **HIGH** | `text_extraction.py:355` | Change `do_table_structure: bool = False` → `True` in both `_build_docling_format_options` and `_text_extraction_pool_initializer` to preserve existing behavior. |
| 2 | **HIGH** | `utils.py:77` | Restore `max_threads: int = 10` — re-add type annotation and default. Public API (`__all__`) cannot drop a defaulted parameter without a deprecation cycle. |
| 3 | **HIGH** | `search_space_preparation.py:272` | After the filter, raise `ValueError` if the result is empty; add a test for the empty-intersection case. |
| 4 | **MEDIUM** | `utils.py:107` | Re-add `int` type annotation; optionally add a runtime guard before `ThreadPoolExecutor`. |
| 5 | **MEDIUM** | `rag_templates_optimization.py:141` | Define `_METADATA_KEYS = frozenset({...})` as a single source of truth, or restructure the YAML to separate `params:` from metadata. |
| 6 | **MEDIUM** | `test_rag_optimization.py:355` | Restructure `test_max_threads_is_accepted` so execution actually reaches thread-pool creation, or rename the test to reflect what it actually asserts. |
| 7 | **LOW** | `utils.py` | Define `_DEFAULT_MAX_THREADS = 10` and reference it from all four call sites instead of repeating the literal. |
| 8 | **LOW** | `search_space_preparation.py:297` | Unify `_validate_chunking_methods` and `_validate_model_list` into a single parameterised `_validate_string_list`. |
| 9 | **LOW** | `mps.py:117` | Change `kwargs.get("evaluator", ...)` → `kwargs.pop(...)` for consistency with every other kwargs consumer in the class. |
| 10 | **LOW** | `test_search_space_prep.py` | Add a test: `chunking_methods=["recursive"]` with a search space that only contains `["hybrid"]` → expect `ValueError` (requires fix #3 first). |

ai4rag/components/data/text_extraction.py
filip-komarzyniec: All implemented according to the ADR document: https://github.com/LukaszCmielowski/architecture-decision-records/blob/autox_docs_updates/documentation/components/autorag/features/experiment_settings.md

For `preset="speed"` which should be the default option the table structure detection should be disabled.

ai4rag/core/experiment/utils.py
filip-komarzyniec: In `AI4RagExperiment` class I'm initialising this parameter by default (if not provided) to `10`, still. Why do you consider this a breaking change, then? 

```python
class AI4RAGExperiment:
    (...)
    # pylint: disable=too-many-arguments,too-many-positional-arguments
    def __init__(
        (...)
        self.max_threads: int = kwargs.pop("max_threads", 10)
```

Mateusz-Switala: Currently, if someone calls the `query_rag` function (public!) with only the `rag` and `questions` arguments, they will get an error with your changes.

ai4rag/components/data/text_extraction.py
Mateusz-Switala: Ok, but we are now in ai4rag and here the default values can be different, I think. In pipeline definition we can set False or True, depending on chosen preset. However, since the functions/methods are quite new, let's make this change, as it seems more natural.

filip-komarzyniec:
## Review response

### Addressed

| # | Severity | Fix |
|---|----------|-----|
| **3** | HIGH | Upfront validation of provided `chunking_methods` against available methods in the search space — raises `ValueError` listing unsupported methods instead of silently filtering. |
| **4** | MEDIUM | Restored `int` type annotation on `query_rag(max_threads)`. |
| **6** | MEDIUM | Reviewed — kept original test name `test_max_threads_is_accepted`; it correctly asserts the kwarg is accepted by the signature. |
| **9** | LOW | Changed `kwargs.get("evaluator", ...)` → `kwargs.pop(...)` in `ModelsPreSelector.__init__`. |
| **10** | LOW | Added `TestUnsupportedChunkingMethods` — passes `chunking_methods=["semantic"]` against a search space with `["recursive", "hybrid"]`, expects `ValueError`. |

### Not addressed

| # | Severity | Reason |
|---|----------|--------|
| **1** | HIGH | Intentional. ADR specifies `do_table_structure=False` for `preset="speed"` (default). Not a regression. |
| **2** | HIGH | Not a breaking change. All public entry points (`AI4RAGExperiment`, `ModelsPreSelector`, `prepare_search_space_report`, `run_rag_optimization`) default `max_threads` to `10` via their own signatures/kwargs. `query_rag` is an internal function plumbed through these; no external caller invokes it directly without a default. |
| **5** | MEDIUM | Over-engineering for 2 metadata keys. The `.pop()` pattern is clear at this scale. |
| **7** | MEDIUM | Moot after fix #4. Each entry point owns its default via signature/kwargs; callers forward `self.max_threads`. No magic `10` duplication problem. |
| **8** | LOW | The two validators have intentionally different strictness (falsy check vs. `isinstance` + `strip`). Not worth abstracting for 2 call sites. |

Mateusz-Switala:
> Moot after fix https://github.com/IBM/ai4rag/pull/4. Each entry point owns its default via signature/kwargs; callers forward self.max_threads. No magic 10 duplication problem.

The problem is that `10` is used several times, with the same context. The natural way would be to create a constant that will represent default `num_threads`.

ai4rag/core/experiment/experiment.py
```diff
@@ -161,6 +161,7 @@ def __init__(
         )
         self.n_mps_embedding_models = kwargs.pop("n_mps_embedding_models", ModelsPreSelector.DEFAULT_N_EMBEDDING_MODELS)
         self.known_observations: list[dict] | None = kwargs.pop("known_observations", None)
+        self.max_threads: int = kwargs.pop("max_threads", 10)
```
jakub-walaszczyk: In my opinion this does not seem to be the best option. `max_threads` is to general here. If you are targeting this to customize only the chat model inference maybe the name should be adjusted.

What comes to my mind first with such parameter is that it could be better to tune this with environmental variable (e.g. `INFERENCE_MAX_THREADS`). If you wish to keep it like that, maybe we should simply change the name.

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -241,10 +267,25 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
...
+        if unsupported:
+            raise ValueError(
+                f"Unsupported chunking methods: {unsupported!r}. "
+                f"Available methods: {sorted(available)!r}."
+            )
+        verbose_repr["chunking_method"] = chunking_methods
+        _logger.info("Chunking methods constrained to: %s", verbose_repr["chunking_method"])
+
+    if contextual_enrichment_enabled is not None:
+        _logger.info("Contextual enrichment enabled: %s", contextual_enrichment_enabled)
```
jakub-walaszczyk: My recommendation is to remove any search space validation from here to `prepare_search_space_with_ogx` based on the provided constraints with proper validators (maybe not as complex as in the previous times).

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -217,6 +242,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
             foundation_models=search_space._search_space["foundation_model"].values,  # pylint: disable=protected-access
             embedding_models=search_space._search_space["embedding_model"].values,  # pylint: disable=protected-access
             metric=metric,
+            max_threads=max_threads,
```
jakub-walaszczyk: Again this `max_threads` seems to be too general.

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -168,6 +178,20 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         Number of benchmark records sampled for model pre-selection.
     random_seed
         Seed for reproducible sampling.
+    chunking_methods
+        When provided, constrains the ``chunking_method`` dimension of the
+        search space to only these methods (e.g. ``["recursive"]`` or
+        ``["hybrid"]``).  ``None`` uses the platform defaults (both
+        ``"recursive"`` and ``"hybrid"``).
+    contextual_enrichment_enabled
```
jakub-walaszczyk: This is not supported now and this is not used further in the stack. Is it prepared for the future `ContextualRetrieval` implementation?

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -134,6 +141,9 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
     top_k_embedding: int = _DEFAULT_TOP_K_EMBEDDING,
     sample_size: int = _DEFAULT_SAMPLE_SIZE,
     random_seed: int = _DEFAULT_SEED,
+    chunking_methods: list[str] | None = None,
+    contextual_enrichment_enabled: bool | None = None,
+    max_threads: int = 10,
```
jakub-walaszczyk: I am wondering whether it would be a better idea to pass `dict` here with parameters tuning search_space. Reasoning:

1. It will be easier to keep backward compatibility when parameters change
2. We can limit number of arguments passed to this already big function
3. We can pass this dict with settings directly to the `search_space`domain, where I think it will be best handled.

ai4rag/components/optimization/rag_templates_optimization.py
```diff
@@ -132,6 +138,7 @@ def run_rag_optimization(  # pylint: disable=too-many-locals,too-many-arguments,
         search_space_raw = yml.safe_load(f)
 
     detected_language: dict[str, str] | None = search_space_raw.pop("detected_language", None)
+    search_space_raw.pop("contextual_enrichment_enabled", None)
```
jakub-walaszczyk: 👀?

ai4rag/components/optimization/search_space_preparation.py
filip-komarzyniec: yes, can delete that as part of this PR so that it is re-introduced alongside implementation of `ContextualRetrieval`.

ai4rag/components/optimization/search_space_preparation.py
filip-komarzyniec: I think it can be done, sure 🙌🏼

ai4rag/components/optimization/rag_templates_optimization.py
filip-komarzyniec: will delete that as `ContextualRetrieval` is not yet implemented.

ai4rag/core/experiment/experiment.py
filip-komarzyniec: I can change the name to `inference_max_threads`. I'm not really convinced about env variable, though: 
- env variable is not explicitly stated in code so it can be unexpected for the users (func signature is much more user-friendly here), 
- env variable has to be explicitly documented, so complexity rises with more vars introduced, 
- env variable requires additional setup of your environment so complexity rises with more vars and more complex execution environments, e.g. cloud envs, 

wdyt? What's more I cannot see, at the moment, any other benefits of utilising env vars here instead of a kwarg, do you have any?

ai4rag/components/optimization/search_space_preparation.py
filip-komarzyniec: we are to move the components' logic code back to pipelines-components repo (the issue will be opened shortly and is already discussed with @LukaszCmielowski), so this will be taken out of `ai4rag`. 

As I do agree with the validation being moved to `prepare_search_space_with_ogx` module, we should prioritise merging this PR in a decent and working state (which the current validation, imo, satisfies). 

Let me investigate the possibility of moving the validation to search space creation phase, however once it proves too complicated and time-consuming, I'll opt for leaving it as-is and trying to make it better in another PR, what do you saY?

ai4rag/components/optimization/search_space_preparation.py
filip-komarzyniec: resolving, we're already discussing this matter in another comment.

ai4rag/components/optimization/search_space_preparation.py
filip-komarzyniec: currently: 
- `prepare_search_space_with_ogx` accepts a `payload` dictionary consisting of models (we have to build this dict from scratch nonetheless inside component), 
- the `prepare_search_space_with_ogx` returns a `SearchSpaceReport` class instance which accepts keyword arguments, 
- the additional parameters are redistributed across different objects and functions calls

Based on the above it seems that providing the input as dict of arguments does not give us any real benefit now (if any). This code is to be moved to pipelines-components repository so maybe then we can try changing the function's signature and see what it helps us with.

ai4rag/core/experiment/experiment.py
jakub-walaszczyk: Let's keep it as it is

ai4rag/components/optimization/search_space_preparation.py
jakub-walaszczyk: Can be in the next iteration

jakub-walaszczyk (commented):
There is one more conflict to resolve

Mateusz-Switala (approved):
lgtm

PR #83, opened 2026-06-29
fix(assets): merge HPO user prompt rules into Responses export system

LukaszCmielowski:
## Description

HPO optimizes RAG patterns using **Chat Completions** (`system_message_text` + formatted `user_message_text` with inline documents). Exported patterns run on OGX via the **Responses API** + `file_search`, where retrieval context and citations are injected at runtime from `benchmarking/rag/config.yaml`.

Previously, `build_pattern_json()` copied only `system_message_text` into `responses_template.input[system]`, dropping user-template rules (RAG behavior, answer length, language policy) that HPO actually tuned. This PR adds export parity via `build_responses_system_input()` while avoiding duplication of OGX-owned prompt text.

## Motivation

- **Eval / deploy mismatch:** HPO sends grounding, length, and language rules in the user message; export sent a weaker system-only prompt → different behavior at inference.
- **OGX responsibility split:** `context_prompt_params`, `annotation_prompt_params`, and `file_search_params` inject retrieval framing and `<|file-id|>` citations at runtime — export must not repeat or substitute that wording.
- **RHOAIENG-71231:** Complete the Chat Completions → Responses parameter mapping started in the responses_template shape work.

## Changes

- Add **`build_responses_system_input()`** to merge non-redundant static supplements from `user_message_text` into `input[role=system]` (persona, answer length, language policy, family-specific rules).
- Strip runtime slots and OGX-duplicative content: `{reference_documents}`, `{question}`, `[1],[2]` citations, `documents below` grounding, and phrases from OGX `config.yaml` templates — **strip only, no replacement text**.
- Wire **`build_pattern_json()`** to use merged system input; preserve `detected_language` on `generation` before export.
- Keep existing **`responses_template`** API shape: structured `input[]`, `tool_choice: file_search`, `tools[]`, `ranking_options`, `temperature`, `max_output_tokens`, `include`.
- Include **`temperature`** and **`max_completion_tokens`** in experiment generation payload so export can bind generation params.
- Add **21 unit tests** covering export parity, all default model families, legacy prompt patterns, and OGX non-duplication.

## Architecture (3-phase flow)

| Phase | Responsibility |
|-------|----------------|
| **HPO** | Chat Completions with full system + user prompts |
| **Export** | Persona + HPO-only rules in `input[system]`; question placeholder in `input[user]`; `file_search` tool config |
| **OGX runtime** | Chunk presentation, retrieval framing, `<\|file-id\|>` citation instructions |

Reference: [chat_completion_to_responses.md](https://github.com/IBM/architecture-decision-records/blob/main/documentation/components/autorag/features/chat_completion_to_responses.md)

## Testing

- [x] `pytest tests/unit/ai4rag/assets_generator/test_pattern_builder.py` (21 tests)
- [x] `uv run black --check ai4rag/`
- [ ] Functional smoke on OGX rag-benchmarks with `enable_annotations: true` (manual)

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (ADR alignment in `architecture-decision-records`)
- [ ] All checks passing (CI)

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/components/assets_generator/pattern_builder.py
+++ b/ai4rag/components/assets_generator/pattern_builder.py
@@ -87,7 +87,7 @@ _USER_GROUNDING_SKIP_PREFIXES = (
     "Do not use outside knowledge",
     "If the documents do not contain the answer",
 )
-_USER_RAG_SCAFFOLD_PREFIXES = (
+_USER_RAG_GROUNDING_PREFIXES = (
     "You are a specialized Retrieval Augmented Generation",
     "Prioritize correctness and ensure your response is grounded",
 )
@@ -126,15 +126,11 @@ def _is_citation_related_line(line: str) -> bool:
     if not stripped:
         return False
     lower = stripped.lower()
-    if any(stripped.startswith(prefix) for prefix in _OGX_DUPLICATIVE_LINE_PREFIXES if "cite" in prefix.lower()):
+    if any(stripped.startswith(prefix) for prefix in _CITATION_PREFIXES):
         return True
     if any(fragment.lower() in lower for fragment in _HPO_CITATION_FRAGMENTS):
         return True
-    if "[1], [2]" in stripped:
-        return True
-    if "document numbers for every factual claim" in lower:
-        return True
-    return False
+    return any(sub.lower() in lower for sub in _CITATION_SUBSTRINGS)
 
 
 def _filter_ogx_duplicative_sentences(line: str) -> str:
@@ -201,7 +197,13 @@ def _strip_ogx_runtime_instructions(text: str) -> str:
 
 
 def _join_answer_scaffold_blocks(lines: list[str]) -> str:
-    """Group lines into blocks separated when an answer-scaffold line starts."""
+    """Group lines into paragraph blocks, starting a new block when an answer-scaffold line appears.
+
+    Scaffold lines are specifically in the form ``Answer (...)`` — e.g.
+    ``"Answer (max 150 words):"`` — as produced by HPO prompt templates.
+    Other leading text such as ``"Answer:"`` or ``"Response:"`` does NOT
+    trigger a new block.
+    """
     if not lines:
         return ""
 
@@ -218,15 +220,12 @@ def _join_answer_scaffold_blocks(lines: list[str]) -> str:
     return "\n\n".join(blocks)
 
 
-def _should_skip_redundant_user_line(stripped: str, system_has_grounding: bool, system_has_citation: bool) -> bool:
+def _should_skip_redundant_user_line(stripped: str, system_has_grounding: bool) -> bool:
     """Return whether a user-template line duplicates system policy for export."""
     if _is_citation_related_line(stripped):
         return True
-    if system_has_citation and stripped.startswith("You MUST cite sources"):
-        return True
-    # Check both grounding prefixes and RAG scaffold prefixes (non-OGX user supplements)
     return system_has_grounding and any(
-        stripped.startswith(prefix) for prefix in _GROUNDING_PREFIXES + _USER_RAG_SCAFFOLD_PREFIXES
+        stripped.startswith(prefix) for prefix in _GROUNDING_PREFIXES + _USER_RAG_GROUNDING_PREFIXES
     )
 
 
@@ -296,17 +295,13 @@ def _extract_static_user_without_reference_slot(text: str) -> str:
 def _system_has_grounding_policy(system: str) -> bool:
     """Return whether the system prompt already states an explicit document-only grounding rule.
 
-    Matches only explicit "answer ONLY using" / "answer using ONLY" instructions.
+    Uses the same prefix list as sentence-level filtering so that adding a new
+    OGX phrase to ``_GROUNDING_PREFIXES`` automatically covers system detection too.
     Does NOT match descriptive personas like "retrieval-augmented assistant" without
-    an explicit grounding constraint — those are system role definitions, not
-    retrieval policies that would make user grounding instructions redundant.
+    an explicit grounding constraint.
     """
     normalized = system.lower()
-    return (
-        "answer using only the provided documents" in normalized
-        or "answer only using information from the documents" in normalized
-        or "answer only using information from documents" in normalized
-    )
+    return any(prefix.lower() in normalized for prefix in _GROUNDING_PREFIXES)
 
 
 def _filter_static_user_for_responses(system: str, static_user: str) -> str:
@@ -320,12 +315,11 @@ def _filter_static_user_for_responses(system: str, static_user: str) -> str:
         return ""
 
     system_has_grounding = _system_has_grounding_policy(system)
-    system_has_citation = "MUST cite sources" in system
 
     filtered_lines: list[str] = []
     for line in static_user.splitlines():
         stripped = line.strip()
-        if not stripped or _should_skip_redundant_user_line(stripped, system_has_grounding, system_has_citation):
+        if not stripped or _should_skip_redundant_user_line(stripped, system_has_grounding):
             continue
         filtered_lines.append(stripped)
 
--- a/ai4rag/components/assets_generator/pattern_builder.py
+++ b/ai4rag/components/assets_generator/pattern_builder.py
@@ -116,12 +116,7 @@ def _line_is_ogx_duplicative(line: str) -> bool:
 
 def _normalize_answer_scaffold(line: str) -> str:
     """Drop citation hints from answer scaffolds; OGX owns citation via annotations."""
-    return (
-        line.replace(", with citations", "")
-        .replace("with citations", "")
-        .replace("  ", " ")
-        .strip()
-    )
+    return line.replace(", with citations", "").replace("with citations", "").replace("  ", " ").strip()
 
 
 def _strip_ogx_runtime_instructions(text: str) -> str:
@@ -170,10 +165,7 @@ def _strip_citation_instructions(text: str) -> str:
 def _system_has_grounding_policy(system: str) -> bool:
     """Return whether the system prompt already states a document-only grounding policy."""
     normalized = system.lower()
-    return (
-        "answer using only the provided documents" in normalized
-        or "retrieval-augmented assistant" in normalized
-    )
+    return "answer using only the provided documents" in normalized or "retrieval-augmented assistant" in normalized
 
 
 def _filter_static_user_for_responses(system: str, static_user: str) -> str:
--- a/ai4rag/components/assets_generator/pattern_builder.py
+++ b/ai4rag/components/assets_generator/pattern_builder.py
@@ -52,6 +52,21 @@ _SYSTEM_GROUNDING_PHRASES = (
     "Answer using ONLY the provided documents.",
     "Answer using ONLY information from documents retrieved via file search.",
 )
+_GROUNDING_DUPLICATE_PREFIXES = (
+    "Answer ONLY using information from the documents",
+    "Do not use outside knowledge",
+    "If the documents do not contain the answer",
+    "You are a specialized Retrieval Augmented Generation",
+    "Prioritize correctness and ensure your response is grounded",
+)
+_USER_GROUNDING_SKIP_PREFIXES = (
+    _HPO_GROUNDING_PREFIX,
+    "Do not use outside knowledge",
+    "If the documents do not contain the answer",
+)
+_DOCUMENT_LABELS = ("Documents:", "Context:", "[Document]")
+_QUESTION_PREFIXES = ("Question:", "Q:", "[conversation]:")
+_LEGACY_DOCUMENT_MARKERS = ("Documents:\n", "Context:\n", "[Document]\n")
 
 
 def _sentence_is_ogx_duplicative(sentence: str) -> bool:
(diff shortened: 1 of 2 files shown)


Mateusz-Switala (commented):
# Code Review: `RHOAIENG-71231-responses-export-system-input-merge`

**Branch:** `RHOAIENG-71231-responses-export-system-input-merge`
**Base:** `remotes/origin/RHOAIENG-71231-define-the-proper-mapping-of-previous-chat-completions-requests-parameters-to-responses-requests-ones`

---

## 1. High-Level Summary

...
---

### 2.1 Dead Code: `_is_citation_line` and `_strip_citation_instructions`

**File:** `pattern_builder.py:170–177`

```python
def _is_citation_line(line: str) -> bool:
    return _line_is_ogx_duplicative(line) and "cite" in line.lower()

def _strip_citation_instructions(text: str) -> str:
    return _strip_ogx_runtime_instructions(text)
```

Neither function is called anywhere in the module. `_strip_citation_instructions` is a pure alias with no differentiated behavior. Both are dead code that will mislead future contributors.

**Severity: Low**

---

### 2.2 Double Computation in `_strip_ogx_runtime_instructions`

**File:** `pattern_builder.py:152–162`

```python
if _line_is_ogx_duplicative(stripped) or _is_citation_related_line(stripped):
    continue  # _line_is_ogx_duplicative calls _filter_ogx_duplicative_sentences internally

cleaned = _filter_ogx_duplicative_sentences(stripped)  # called again on the same `stripped`
```

For every line that passes the first guard (i.e., is *not* duplicative), `_filter_ogx_duplicative_sentences` is called a second time. `_line_is_ogx_duplicative` already invokes `_filter_ogx_duplicative_sentences` internally and discards the result. This is wasted work — harmless for small prompt strings, but conceptually wrong: the function pattern implies `_line_is_ogx_duplicative` is a boolean check, not a filtering step, yet it gates a redundant second call that does the real work.

**Severity: Low**

---

### 2.3 Redundant Variable `raw_system` in `build_responses_system_input`

**File:** `pattern_builder.py:351–353`

```python
system = _adapt_system_for_responses_export((generation.get("system_message_text") or "").strip())
user_template = generation.get("user_message_text") or ""
raw_system = (generation.get("system_message_text") or "").strip()  # identical expression to line above
```

`raw_system` is computed with an identical expression to what feeds `system`. The intent is correct — pass unprocessed text to `_filter_static_user_for_responses` so the grounding heuristic operates on the original content — but the implicit coupling is fragile. If the first line is refactored, `raw_system` silently diverges. The asymmetry is also undocumented at the call site.

**Severity: Low**

---

### 2.4 Over-Broad Grounding Heuristic in `_system_has_grounding_policy`

**File:** `pattern_builder.py:270–273`

```python
def _system_has_grounding_policy(system: str) -> bool:
    normalized = system.lower()
    return "answer using only the provided documents" in normalized or "retrieval-augmented assistant" in normalized
```

The second condition matches any system prompt mentioning "retrieval-augmented assistant" — a persona label — and treats it as a document-only grounding *policy*. A prompt such as `"You are a retrieval-augmented assistant. Use your best judgment."` would incorrectly trigger the heuristic and suppress user-template lines starting with `_GROUNDING_DUPLICATE_PREFIXES` even when those lines are needed.

Additionally, the check runs against `raw_system` (the unprocessed original). If grounding phrases were stripped from `system` by `_adapt_system_for_responses_export`, user lines that repeat those phrases should arguably be *kept* (the system no longer states the policy), but the current logic still suppresses them.

**Severity: Medium**

---

### 2.5 `ranker_alpha != 1` Condition Silently Falls Through for Alpha=1.0

**File:** `pattern_builder.py:429`

```python
elif search_mode == "hybrid" and ranker_strategy == "weighted" and ranker_alpha is not None and ranker_alpha != 1:
```

If a caller explicitly sets `ranker_strategy = "weighted"` and `ranker_alpha = 1.0` on a hybrid search pattern (semantically meaning vector-only within hybrid mode), the `elif` branch is skipped and the `else` branch fires — producing the correct output `{"ranker": "weighted", "alpha": 1.0}` by coincidence. The behavior is correct but reached via the wrong logical branch, and there is no test covering this case.

**Severity: Low**

---

### 2.6 Interior Double-Space After `_SYSTEM_GROUNDING_PHRASES` Substitution

**File:** `pattern_builder.py:142–144`

```python
for phrase in _SYSTEM_GROUNDING_PHRASES:
    text = text.replace(phrase, "").replace(phrase.rstrip("."), "")
```

When a phrase appears mid-sentence — e.g., `"You are a helpful assistant. Answer using ONLY the provided documents. Be concise."` — after replacement the text becomes `"You are a helpful assistant.  Be concise."` (double space). The final `.strip()` on line 167 only removes leading/trailing whitespace, not interior double-spaces. `_normalize_answer_scaffold` replaces `"  "` with `" "` but is only applied per-line after other filters, not at this stage.

**Severity: Low**

---

### 3.1 Overlapping and Partially Redundant Prefix Tuples

The module defines five distinct prefix/substring lists with overlapping content and subtle semantic differences not documented at their definitions:

| Constant | Entries | Used In |
|---|---|---|
| `_OGX_DUPLICATIVE_LINE_PREFIXES` | 14 prefixes | `_sentence_is_ogx_duplicative`, `_is_citation_related_line` |
| `_OGX_DUPLICATIVE_SUBSTRINGS` | 8 substrings | `_sentence_is_ogx_duplicative` |
| `_GROUNDING_DUPLICATE_PREFIXES` | 5 prefixes | `_should_skip_redundant_user_line` |
| `_USER_GROUNDING_SKIP_PREFIXES` | 3 prefixes | `_should_skip_user_export_line` |
| `_HPO_CITATION_FRAGMENTS` | 3 strings | `_is_citation_related_line`, `_strip_ogx_runtime_instructions` |

`"Answer ONLY using information from the documents"` appears in both `_OGX_DUPLICATIVE_LINE_PREFIXES` and `_GROUNDING_DUPLICATE_PREFIXES`. `"Do not use outside knowledge"` appears in both `_OGX_DUPLICATIVE_LINE_PREFIXES` and `_USER_GROUNDING_SKIP_PREFIXES`. A contributor adding a new OGX-owned phrase must know which of five lists to update.

**Severity: Medium**

---

### 3.2 `_DOCUMENT_SLOT_MARKERS` vs `_DOCUMENT_LABELS` Parallelism Is Confusing

```python
_DOCUMENT_SLOT_MARKERS = frozenset({"[Document]", "[End]", "Documents:", "Context:"})
_DOCUMENT_LABELS = ("Documents:", "Context:", "[Document]")
```

These share three members but differ in type (frozenset vs tuple), inclusion (`[End]` only in frozenset), and usage context. The asymmetry around `[End]` is intentional — `[End]` follows the document slot and is handled by the suffix extractor, while labels before the slot use `_DOCUMENT_LABELS` — but nothing in either definition explains this split.

**Severity: Low**

---

### 3.3 Two-Pass User Content Filtering Without Explanation

**File:** `pattern_builder.py:355–360`

```python
static_user = _adapt_static_user_for_responses_export(
    _filter_static_user_for_responses(
        raw_system,
        _extract_static_user_instructions(user_template),
    ),
)
```

Two successive filter passes over user content use two different predicate sets (`_should_skip_redundant_user_line` vs `_should_skip_user_export_line`). Neither the caller nor the intermediate functions document the expected ordering or why two passes are necessary instead of one combined predicate. A reader must trace through four functions to understand what survives.

**Severity: Medium**

---

### 3.4 `_normalize_answer_scaffold` Single-Pass Double-Space Removal

**File:** `pattern_builder.py:132–134`

```python
return line.replace(", with citations", "").replace("with citations", "").replace("  ", " ").strip()
```

The `"  " → " "` replacement is a single pass. Adding another `replace()` call in the future could introduce triple-spaces that a single pass would not fully collapse. The pattern is also undocumented — the comment explains the *why* but not the fragility.

**Severity: Low**

---

### 3.5 String-Matching Approach Creates Invisible Coupling to OGX Config

The entire filtering strategy relies on exact prefix/substring matches against phrases defined in OGX's `benchmarking/rag/config.yaml`. There is a code comment referencing this file, but no mechanism to fail loudly if OGX changes its injection strings. If OGX renames any phrase, the export will silently include duplicative content and all existing tests will still pass because they only test known-good inputs.

**Severity: Medium** (architectural)

---

## 4. Test Coverage Gaps

### 4.1 `test_build_pattern_json_uses_export_parity_system_input` Is Tautological

**File:** `test_pattern_builder.py:222–242`

```python
build_pattern_json(pattern)

generation = pattern["settings"]["generation"]
expected = build_responses_system_input(generation)   # called AFTER build_pattern_json mutated pattern
actual = pattern["settings"]["responses_template"]["input"][0]["content"][0]["text"]
assert actual == expected                             # asserting f(x) == f(x)
```

`build_pattern_json` internally calls `build_responses_system_input(generation)` and writes the result into `responses_template`. The test then calls `build_responses_system_input(generation)` again on the same unchanged `generation` and asserts they're equal. This always passes by construction. The useful assertions are the specific content checks below (`"Granite Chat" in actual`, etc.); the `assert actual == expected` line adds no signal.

**Severity: Medium**

---

### 4.2 No Test for `hybrid + weighted + alpha=1.0`

The condition `ranker_alpha != 1` on line 429 means a hybrid-weighted pattern with `alpha=1.0` falls through to the `else` branch. The output is correct but the logical path is wrong, and there is no test documenting this edge case or validating the behavior is intentional.

**Severity: Medium**

---

### 4.3 No Test for Missing Required Keys in `build_pattern_json`

`build_pattern_json` directly accesses `generation["model_id"]`, `generation["max_completion_tokens"]`, and `generation["temperature"]` without `.get()` guards. If any field is absent, a `KeyError` is raised. `build_responses_system_input`, by contrast, uses `.get()` defensively throughout. The inconsistency is undocumented and untested — there is no test verifying the error type or message, nor any validation at function entry.

**Severity: Medium**

---

### 4.4 No Test for Empty/None System and User Prompts Together

`build_responses_system_input` with both `system_message_text=""` and `user_message_text=""` returns `""`. There is no test asserting this baseline, nor a test for when only one input is empty while the other is non-empty.

**Severity: Low**

---

### 4.5 No Test for Mixed-Content System Prompt Sentence Filtering

There is no test asserting that a system prompt where only *some* sentences are OGX-duplicative has those sentences correctly excised while surrounding sentences are preserved. For example, `"You are an expert assistant. Answer using ONLY the provided documents. Be concise."` should produce `"You are an expert assistant. Be concise."` — but this path is not explicitly covered.

**Severity: Medium**

---

### 4.6 No Test for User Template Without Any Slot

`_extract_static_user_instructions` handles templates with `{reference_documents}` and templates with legacy document markers. A user template that is pure static text (no slots, no markers) falls through to `_extract_static_user_without_reference_slot`, which returns the entire text unchanged. This path has no dedicated test.

**Severity: Low**

---

## 5. Actionable Fixes

### Fix 1 — Remove Dead Code

**Rationale:** `_is_citation_line` and `_strip_citation_instructions` are never called. Dead functions add cognitive load and mislead future contributors.
**Severity: Low**

```python
# DELETE both functions entirely:
def _is_citation_line(line: str) -> bool: ...
def _strip_citation_instructions(text: str) -> str: ...
```

---

### Fix 2 — Eliminate Double Call to `_filter_ogx_duplicative_sentences`

**Rationale:** The function is called inside `_line_is_ogx_duplicative` and then again on the same input. Cache the result instead.
**Severity: Low**

```python
# Replace (in _strip_ogx_runtime_instructions):
if _line_is_ogx_duplicative(stripped) or _is_citation_related_line(stripped):
    continue
cleaned = _filter_ogx_duplicative_sentences(stripped)

# With:
cleaned = _filter_ogx_duplicative_sentences(stripped)
if not cleaned or _is_citation_related_line(stripped):
    continue
```

---

### Fix 3 — Rename `raw_system` and Document the Intent

**Rationale:** The dual-variable pattern is correct but opaque to readers without explicit documentation.
**Severity: Low**

```python
def build_responses_system_input(generation: dict) -> str:
    original_system = (generation.get("system_message_text") or "").strip()
    # original_system is kept separately so the grounding-dedup heuristics
    # operate on the author's intent, not on the stripped output.
    exported_system = _adapt_system_for_responses_export(original_system)
    user_template = generation.get("user_message_text") or ""

    static_user = _adapt_static_user_for_responses_export(
        _filter_static_user_for_responses(
            original_system,
            _extract_static_user_instructions(user_template),
        ),
    )
    ...
```

---

### Fix 4 — Tighten `_system_has_grounding_policy` Heuristic

**Rationale:** `"retrieval-augmented assistant"` is a persona descriptor, not evidence of a document-only grounding policy.
**Severity: Medium**

```python
def _system_has_grounding_policy(system: str) -> bool:
    normalized = system.lower()
    return (
        "answer using only the provided documents" in normalized
        or "answer only using information from" in normalized
        or "retrieval augmented generation (rag) assistant" in normalized  # full phrase only
    )
```

Remove the partial `"retrieval-augmented assistant"` match and replace it with a phrase that explicitly describes RAG document-restriction behavior.

---

### Fix 5 — Add Test for `hybrid + weighted + alpha=1.0`

**Rationale:** The `elif ranker_alpha != 1` guard causes this combination to hit the `else` branch silently.
**Severity: Medium**

```python
def test_hybrid_weighted_alpha_one_falls_to_semantic_simulation(self):
    """Hybrid weighted with alpha=1.0 produces the same output as the default semantic-only case."""
    pattern = _make_pattern()
    pattern["settings"]["retrieval"]["search_mode"] = "hybrid"
    pattern["settings"]["retrieval"]["ranker_strategy"] = "weighted"
    pattern["settings"]["retrieval"]["ranker_alpha"] = 1.0

    build_pattern_json(pattern)

    ro = pattern["settings"]["responses_template"]["tools"][0]["ranking_options"]
    assert ro == {"ranker": "weighted", "alpha": 1.0}
```

Also consider changing the `elif` condition to `ranker_alpha != 1.0` (explicit float) and/or adding a separate `elif` for the `alpha=1.0` hybrid case that falls through deliberately to the default — making the intent visible rather than relying on guard-clause fall-through.

---

### Fix 6 — Fix Tautological Test

**Rationale:** `assert actual == expected` where both values are produced by calling `build_responses_system_input` on the same input is a no-op assertion.
**Severity: Medium**

```python
# Capture expected BEFORE calling build_pattern_json:
expected = build_responses_system_input({
    "system_message_text": get_system_message_text(model_id),
    "user_message_text": get_user_message_text(model_id, language_autodetect=False),
})

pattern = _make_pattern()
pattern["settings"]["generation"]["model_id"] = model_id
pattern["settings"]["generation"]["system_message_text"] = get_system_message_text(model_id)
pattern["settings"]["generation"]["user_message_text"] = get_user_message_text(model_id, language_autodetect=False)

build_pattern_json(pattern)

actual = pattern["settings"]["responses_template"]["input"][0]["content"][0]["text"]
assert actual == expected  # now genuinely cross-checks build_pattern_json against build_responses_system_input
assert actual != pattern["settings"]["generation"]["system_message_text"]
...
```

---

### Fix 7 — Add Test for Mixed-Content System Prompt Sentence Filtering

**Rationale:** No test covers the partial-line filtering path where one sentence in a multi-sentence system prompt is OGX-duplicative.
**Severity: Medium**

```python
def test_strip_ogx_runtime_instructions_partial_sentence_removal(self):
    """OGX sentences in a multi-sentence system prompt are excised; others are preserved."""
    generation = {
        "system_message_text": (
            "You are an expert assistant. "
            "Answer using ONLY the provided documents. "
            "Be concise."
        ),
        "user_message_text": "",
    }
    result = build_responses_system_input(generation)
    assert "You are an expert assistant" in result
    assert "Be concise" in result
    assert "provided documents" not in result.lower()
```

---

### Fix 8 — Consolidate Overlapping Prefix Lists

**Rationale:** Five overlapping lists are hard to audit and extend. A new OGX phrase requires manual identification of which lists to update.
**Severity: Medium**

Define a single authoritative collection of OGX-owned phrases and derive role-specific views from it:

```python
# All phrases OGX injects at file_search runtime.
# Source: benchmarking/rag/config.yaml (file_search_params, context_prompt_params,
# annotation_prompt_params). Update this set when OGX config changes.
_OGX_RUNTIME_LINE_PREFIXES: tuple[str, ...] = (
    "You MUST cite sources",
    "Cite sources immediately",
    "Answer ONLY using information from the documents",
    # ... full list
)

# Subset used for system-vs-user deduplication — grounding statements only.
_GROUNDING_DUPLICATE_PREFIXES = tuple(
    p for p in _OGX_RUNTIME_LINE_PREFIXES
    if p.startswith(("Answer ONLY", "Do not use", "If the documents", "You are a specialized"))
)
```

---

### Fix 9 — Guard Against Missing Required Keys in `build_pattern_json`

**Rationale:** `build_responses_system_input` uses `.get()` defensively; `build_pattern_json` uses direct key access. The inconsistency creates a hidden contract violation.
**Severity: Medium**

Add entry-point validation:

```python
def build_pattern_json(pattern: dict, detected_language: dict | None = None) -> dict:
    generation = pattern["settings"]["generation"]
    required_keys = {"model_id", "max_completion_tokens", "temperature"}
    missing = required_keys - generation.keys()
    if missing:
        raise ValueError(f"generation dict is missing required fields: {missing}")
    ...
```

Alternatively, document the expected schema in the docstring and explicitly state that `KeyError` is the caller's responsibility.

PR #84, opened 2026-07-01
fix(embedding): truncate chunks exceeding context length in OGXEmbeddingModel

jakub-walaszczyk:
## Description

Adds a two-layer defense against `BadRequestError` when embedding chunks that exceed the model's context length in `OGXEmbeddingModel`. The first layer passes `truncate_prompt_tokens: -1` to the vLLM server so it truncates inputs server-side. The second layer is a client-side fallback that progressively truncates oversized texts using character-based token estimates when the server cannot handle truncation on its own.

## Motivation

When chunks exceed the embedding model's `context_length`, the OGX/vLLM server rejects the entire batch with a `BadRequestError`. This causes indexing pipelines to fail on corpora containing long documents. Server-side truncation via `truncate_prompt_tokens` is the preferred fix, but not all vLLM deployments support this parameter — hence the client-side fallback is needed as a safety net.

## Changes

| Area | What changed |
|------|-------------|
| `ai4rag/rag/embedding/ogx.py` | Extracted `_call_embedding_api` from `_embed_text` to isolate the raw API call and add `truncate_prompt_tokens: -1` to the request body |
| `ai4rag/rag/embedding/ogx.py` | Added `_truncate_text` helper that clips text to `context_length * (1 - margin) * _CHARS_PER_TOKEN` characters |
| `ai4rag/rag/embedding/ogx.py` | Rewrote `_embed_text` as an orchestrator: try the batch as-is, then retry the entire batch with 5% and 10% truncation margins on `BadRequestError` |
| `ai4rag/rag/embedding/ogx.py` | Added module constants `_CHARS_PER_TOKEN = 5` and `_TRUNCATION_MARGINS = (0.05, 0.10)` |
| `tests/unit/ai4rag/rag/embedding/test_ogx.py` | Added `TestEmbeddingTruncationFallback` suite (9 tests) covering: no-fallback success, first/second margin retry, all-retries-fail raises, short texts unaffected, document order preserved, and `embed_query` fallback |

## Testing

- [x] `pytest tests/unit/ai4rag/rag/embedding/test_ogx.py` (9 new + existing tests)
- [ ] Functional smoke: index a corpus with chunks near the context boundary and verify no API failures

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/rag/embedding/ogx.py
+++ b/ai4rag/rag/embedding/ogx.py
@@ -127,7 +127,9 @@ class OGXEmbeddingModel(BaseEmbeddingModel[OgxClient, OGXEmbeddingParams]):
 
         return [
             data.embedding
-            for data in self.client.embeddings.create(input=text_input, model=self.model_id).data
+            for data in self.client.embeddings.create(
+                input=text_input, model=self.model_id, extra_body={"truncate_prompt_tokens": -1}
+            ).data
             if not isinstance(data.embedding, str)
         ]
 
--- a/ai4rag/rag/embedding/ogx.py
+++ b/ai4rag/rag/embedding/ogx.py
@@ -15,7 +15,7 @@ __all__ = ["OGXEmbeddingModel", "OGXEmbeddingParams"]
 
 
 MIN_CONTEXT_LENGTH = 700
-_CHARS_PER_TOKEN = 5
+_CHARS_PER_TOKEN = 3
 _TRUNCATION_MARGINS = (0.05, 0.10)
 
 
@@ -93,6 +93,11 @@ class OGXEmbeddingModel(BaseEmbeddingModel[OgxClient, OGXEmbeddingParams]):
         remaining interval is smaller than 256 tokens, keeping the number
         of API calls to roughly 5.
 
+        Calls the embedding API directly with ``truncate_prompt_tokens``
+        set to ``None`` so that oversized probes are rejected by the
+        server rather than silently truncated (the server may default
+        to ``-1`` which auto-truncates to ``max_model_len``).
+
         Raises
         ------
         RuntimeError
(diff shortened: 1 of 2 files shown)


ai4rag/rag/embedding/ogx.py
```diff
@@ -123,13 +192,85 @@ def _embed_text(self, text_input: list[str] | str) -> list[list[float]]:
...
+            max_tokens = int(self._params.context_length * _PROACTIVE_CONTEXT_RATIO)
+            text_input = self._truncate_texts(text_input, max_tokens)
 
-        return [
-            data.embedding
-            for data in self.client.embeddings.create(input=text_input, model=self.model_id).data
-            if not isinstance(data.embedding, str)
-        ]
+        try:
+            response = self.client.embeddings.create(input=text_input, model=self.model_id)
```
MichalSteczko:
```python
self.client.embeddings.create(input=text_input, model=self.model_id, metadata={"extra_body": {"truncate_prompt_tokens": -1}})
```

ai4rag/rag/embedding/ogx.py
```diff
@@ -111,8 +128,60 @@ def _detect_context_length(self) -> int:
...
+            tokens = self._tiktoken_encoding.encode(text)
+            if len(tokens) <= max_tokens:
+                return text
+            logger.warning(
+                "Truncation for model '%s': %d -> %d tokens",
+                self.model_id,
+                len(tokens),
+                max_tokens,
+            )
+            return self._tiktoken_encoding.decode(tokens[:max_tokens])
```
MichalSteczko: Are we sure that we want to this like that?
Shouldn't we do it on vLLM side with `extra_body` passed to `embedding` request like shown in the comment: https://github.com/IBM/ai4rag/pull/84/changes#r3506130528.

MichalSteczko (approved):

PR #85, opened 2026-07-01
fix(hpo): wire random_state parameter for deterministic optimization

### LINKED ISSUE #76: [Bug]: Title: GAMOptSettings.random_state is unused — GAM exploration is non-deterministic
### Bug Description

# GAMOptSettings.random_state is unused — GAM exploration order is non-deterministic

## Summary

`GAMOptSettings` exposes `random_state` (default `64`), but `GAMOptimizer` never uses it. Initial hyperparameter exploration calls `random.shuffle()` on the global RNG, so repeated experiments with the same settings can evaluate **different RAG config sequences** (retrieval window, chunk count, etc.).

This blocks reproducible benchmarks and confounds prompt A/B tests (e.g. baseline vs PR #75) when GAM is enabled.

## Root cause

In `ai4rag/core/hpo/gam_opt.py`:

```python
@dataclass
class GAMOptSettings(OptimizerSettings):
    n_random_nodes: int = 4
    evals_per_trial: int = 1
    random_state: int = 64   # defined but unused
```

```python
def evaluate_initial_random_nodes(self) -> None:
    ...
    random.shuffle(combinations_local)  # global RNG, ignores random_state
```

`RandomOptimizer` (`random_opt.py`) has the same pattern — no seed parameter, uses `random.shuffle(combinations)`.

Docs list `random_state`: https://ibm.github.io/ai4rag/latest/api-reference/core/hpo/

### 1. Wire `random_state` in GAM (and Random) optimizers

```python
# GAMOptimizer.__init__
self._rng = random.Random(self.settings.random_state)

# evaluate_initial_random_nodes
self._rng.shuffle(combinations_local)
```

Add `random_state: int = 64` to `RandomOptSettings` and use the same pattern in `RandomOptimizer`.

### 2. Optional: experiment-level seed

`AI4RAGExperiment(..., experiment_seed: int | None = None)` — if set, propagate to optimizer `random_state` when not explicitly provided.

### 3. Tests

- Mock objective with deterministic scores.
- Assert two `GAMOptimizer.search()` runs with identical `search_space`, `GAMOptSettings(max_evals=N, n_random_nodes=M, random_state=42)` evaluate the **same ordered list** of param dicts.

### 4. Docs

Update `user-guide/optimizers.md` with reproducibility guidance.

## Related improvements (separate / follow-up)

| Item | Why |
|------|-----|
| Benchmark mode `temperature=0` | Reduce LLM score variance during eval |
| `evaluate_fixed_patterns(patterns: list[dict])` or document `known_observations` for replay | Prompt-only A/B without GAM confounding |
| Persist exploration trace (evaluated params + scores per step) | Debug cross-run differences |

## References

- HPO API: https://ibm.github.io/ai4rag/latest/api-reference/core/hpo/
- Comparative analysis: RHOAI/AutoRAG William benchmark notebooks using `GAMOptSettings(max_evals=10, n_random_nodes=4)`

## Environment

- ai4rag: `move-autorag-components-code-to-ai4rag` / `fix-prompts` branches
- Optimizer: `GAMOptimizer` via `AI4RAGExperiment.search()`
- Metric: `faithfulness` (Unitxt)

### Steps to Reproduce

Run GAM based experiment multiple times.

## Expected behavior

- `GAMOptSettings(random_state=42)` → same sequence of evaluated param dicts across runs (given deterministic objective scores).
- Document what `random_state` controls (config exploration order) vs what it does **not** control (LLM stochasticity at `temperature=0.2`).

## Observed behavior

William benchmark comparative runs (`max_evals=10`, `n_random_nodes=4`) produced different “best” retrieval configs across runs on the **same** ai4rag branch — e.g. baseline best `window=3, chunks=5` vs PR #75 best `window=1, chunks=3`. `Pattern1`…`Pattern10` labels are evaluation ordinals, not fixed configs, so cross-run pattern names are not comparable.

### Environment

```markdown
- ai4rag version:
- Python version:
- OGX version:
- Vector Store (Milvus/ChromaDB):
- Operating System:
```

### Error Logs

```shell

```

### Additional Context

_No response_

LukaszCmielowski:
Fixes #76

Both GAMOptimizer and RandomOptimizer now use dedicated RNG instances seeded with settings.random_state instead of the global random module. This enables reproducible HPO experiments with identical exploration order across runs.

Changes:
- GAMOptimizer: Add self._rng instance, use in evaluate_initial_random_nodes
- RandomOptimizer: Add random_state to settings, add self._rng instance, use in search
- RandomOptimizer: Copy combinations list before shuffle to avoid mutating original
- Add 4 new determinism tests (2 per optimizer) verifying same/different random_state behavior
- Update test_random_opt_settings_to_dict to expect random_state in output
- Update test_search_shuffles_combinations to mock optimizer._rng instead of global random

All 53 optimizer tests passing (37 GAM + 16 Random).

## Description
Brief summary of changes

## Motivation
Why is this change needed?

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/core/hpo/base_optimizer.py
+++ b/ai4rag/core/hpo/base_optimizer.py
@@ -28,6 +28,12 @@ class OptimizerSettings:
     ----------
     max_evals : int
         Maximum number of evaluations performed during optimization process.
+    random_state : int, default=64
+        Seed for random number generator controlling exploration order.
+        Use the same seed across runs to get deterministic evaluation
+        sequences (given deterministic objective function scores).
+        Does NOT control: LLM temperature, benchmark sampling, or other
+        non-optimizer randomness.
 
     Methods
     -------
@@ -36,6 +42,7 @@ class OptimizerSettings:
     """
 
     max_evals: int
+    random_state: int = 64
 
     def to_dict(self) -> dict:
         """
--- a/ai4rag/core/hpo/base_optimizer.py
+++ b/ai4rag/core/hpo/base_optimizer.py
@@ -32,8 +32,6 @@ class OptimizerSettings:
         Seed for random number generator controlling exploration order.
         Use the same seed across runs to get deterministic evaluation
         sequences (given deterministic objective function scores).
-        Does NOT control: LLM temperature, benchmark sampling, or other
-        non-optimizer randomness.
 
     Methods
     -------
--- a/ai4rag/core/hpo/gam_opt.py
+++ b/ai4rag/core/hpo/gam_opt.py
@@ -75,6 +75,7 @@ class GAMOptimizer(BaseOptimizer):
         self.evaluations = []
         self._evaluated_combinations = []
         self._encoders_with_columns: list[tuple[str, LabelEncoder]] = []
+        self._rng = random.Random(self.settings.random_state)
 
         if known_observations:
             self._load_known_observations(known_observations)
@@ -193,7 +194,7 @@ class GAMOptimizer(BaseOptimizer):
             return
 
         combinations_local = [c for c in copy(self._search_space.combinations) if c not in self._evaluated_combinations]
-        random.shuffle(combinations_local)
+        self._rng.shuffle(combinations_local)
 
         gen = (x for x in combinations_local)
 
--- a/ai4rag/core/hpo/gam_opt.py
+++ b/ai4rag/core/hpo/gam_opt.py
@@ -29,6 +29,8 @@ class GAMOptSettings(OptimizerSettings):
 
     Parameters
     ----------
+    max_evals : int
+        Maximum number of evaluations performed during optimization process.
     n_random_nodes : int, default=4
         Number of random configurations to evaluate before starting GAM iterations.
     evals_per_trial : int, default=1
@@ -36,7 +38,7 @@ class GAMOptSettings(OptimizerSettings):
     random_state : int, default=64
         Inherited from OptimizerSettings. Controls shuffle order of initial
         random exploration phase. Does NOT control GAM model randomness
-        (GAM training is deterministic) or LLM generation stochasticity.
+        (GAM training is deterministic).
     """
 
     n_random_nodes: int = 4
@@ -85,7 +87,6 @@ class GAMOptimizer(BaseOptimizer):
         self.evaluations = []
         self._evaluated_combinations = []
         self._encoders_with_columns: list[tuple[str, LabelEncoder]] = []
-        self._rng = random.Random(self.settings.random_state)
 
         if known_observations:
             self._load_known_observations(known_observations)
@@ -204,7 +205,7 @@ class GAMOptimizer(BaseOptimizer):
             return
 
         combinations_local = [c for c in copy(self._search_space.combinations) if c not in self._evaluated_combinations]
-        self._rng.shuffle(combinations_local)
+        random.Random(self.settings.random_state).shuffle(combinations_local)
 
         gen = (x for x in combinations_local)
 
--- a/ai4rag/core/hpo/gam_opt.py
+++ b/ai4rag/core/hpo/gam_opt.py
@@ -26,11 +26,21 @@ class GAMOptSettings(OptimizerSettings):
     Settings for the GAMOptimizer. For the detailed description
     of parameters for Generalized Additive Models, please see pygam
     documentation.
+
+    Parameters
+    ----------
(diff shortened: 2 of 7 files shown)


ai4rag/core/hpo/base_optimizer.py
```diff
@@ -28,6 +28,12 @@ class OptimizerSettings:
     ----------
     max_evals : int
         Maximum number of evaluations performed during optimization process.
+    random_state : int, default=64
+        Seed for random number generator controlling exploration order.
+        Use the same seed across runs to get deterministic evaluation
+        sequences (given deterministic objective function scores).
+        Does NOT control: LLM temperature, benchmark sampling, or other
+        non-optimizer randomness.
```
jakub-walaszczyk: I do not think we need to introduce this context to the optimizer settings. Potentially optimizer classes can be taken out and reused in other place. I think we should remove these 2 lines of what it does not do.

ai4rag/core/hpo/gam_opt.py
```diff
@@ -26,11 +26,21 @@ class GAMOptSettings(OptimizerSettings):
...
+    Parameters
+    ----------
+    n_random_nodes : int, default=4
+        Number of random configurations to evaluate before starting GAM iterations.
+    evals_per_trial : int, default=1
+        Number of configurations to evaluate per GAM iteration.
+    random_state : int, default=64
+        Inherited from OptimizerSettings. Controls shuffle order of initial
+        random exploration phase. Does NOT control GAM model randomness
+        (GAM training is deterministic) or LLM generation stochasticity.
```
jakub-walaszczyk: Same here, to remove that it does not control LLM. GAM model randomness is fair and accurate to keep.

ai4rag/core/hpo/gam_opt.py
```diff
@@ -75,6 +85,7 @@ def __init__(
         self.evaluations = []
         self._evaluated_combinations = []
         self._encoders_with_columns: list[tuple[str, LabelEncoder]] = []
+        self._rng = random.Random(self.settings.random_state)
```
jakub-walaszczyk: We do not need to store it in the instance, do we?

ai4rag/core/hpo/gam_opt.py
```diff
@@ -193,7 +204,7 @@ def evaluate_initial_random_nodes(self) -> None:
             return
 
         combinations_local = [c for c in copy(self._search_space.combinations) if c not in self._evaluated_combinations]
-        random.shuffle(combinations_local)
+        self._rng.shuffle(combinations_local)
```
jakub-walaszczyk: Instead of storing `_rng` maybe we could simply: `random.Random(self.settings.random_state).shuffle(combinations_local)`

ai4rag/core/hpo/random_opt.py
```diff
@@ -15,7 +15,17 @@
...
+    """Settings for random optimizer.
+
+    Parameters
+    ----------
+    max_evals : int
+        Maximum number of configurations to evaluate.
+    random_state : int, default=64
+        Inherited from OptimizerSettings. Controls shuffle order of
+        search space combinations. Does NOT control LLM generation
+        stochasticity or benchmark sampling randomness.
```
jakub-walaszczyk: Same story: maybe we should remove part about LLM generation settings here.

jakub-walaszczyk:
Why removing `self._rng` is the right call:

  1. It was a footgun. A stored `random.Random` instance carries mutable state — after `shuffle()`, its internal position
  advances. If `search()` were called twice on the same optimizer (or if any other method used `self._rng`), the second
  call would produce different results with the same seed. That violates the contract implied by `random_state`.
  2. Inline `random.Random(seed).shuffle(...)` is idempotent. Each call creates a fresh RNG from the seed, so identical
  inputs always produce identical outputs regardless of how many times `search()` runs. This is the correct pattern for
  seed-controlled reproducibility.
  3. No shared state across subclasses. `RandomOptimizer` and `GAMOptimizer` each shuffle independently in their own `search()/evaluate_initial_random_nodes().` A base-class `_rng` would have implied a shared sequencing contract that doesn't exist.
  4. Minimal allocation cost. Constructing `random.Random(int)` is trivially cheap — there's no performance reason to
  cache it.

jakub-walaszczyk (approved):

PR #89, opened 2026-07-02
chore: add validation for chunk sizes and chunking methods

Mateusz-Switala:
Assisted-by: Claude Code

## Description

  Extends `prepare_search_space_with_ogx` to accept optional `chunking_method` and `chunk_size`
  overrides directly via the payload, and adds `chunk_sizes` as a new parameter to
  `prepare_search_space_report`. Also lowers the minimum allowed chunk size to 128 tokens and
  deduplicates any repeated values in the chunking inputs.

  ## Motivation

  The existing `prepare_search_space_with_ogx` always used platform-default chunking methods
  (`recursive`, `hybrid`) and chunk sizes (`512`, `1024`, `2048`). Users running experiments on
  shorter documents or with a single chunking strategy had no way to constrain these dimensions
  without post-processing the search space manually. This change exposes that control at the
  preparation step, in the single function callers already use.

  ## Changes

  - **`ai4rag/utils/constants.py`** — lowered `ChunkingConstraints.MIN_CHUNK_SIZE` from `512` to `128`
  - **`ai4rag/search_space/prepare/input_payload_types.py`** — added optional `chunking_methods` and
    `chunk_sizes` fields to `AI4RAGConstraints`; both are pydantic-validated (non-empty list,
    non-empty strings / positive integers respectively)
  - **`ai4rag/search_space/prepare/prepare_search_space.py`**
    - Extracted three private helpers to eliminate duplication: `_resolve_models_from_payload`,
      `_apply_language_detection`, `_build_model_params`
    - Extended `prepare_search_space_with_ogx` to accept and validate `chunking_methods` and
      `chunk_sizes` in the payload; when provided they override the platform defaults for those
      dimensions; duplicate values are deduplicated with a warning
    - Updated docstring to clarify that only model validation goes through OGX; chunking is
      validated locally against `ChunkingConstraints`
  - **`ai4rag/search_space/prepare/__init__.py`** — removed `prepare_search_space_custom`
    (functionality merged into `prepare_search_space_with_ogx`)
  - **`ai4rag/components/optimization/search_space_preparation.py`**
    - Added `chunk_sizes: list[int] | None = None` parameter to `prepare_search_space_report`
    - Updated `_validate_chunking_methods` to validate values against `ChunkingConstraints.METHODS`,
      providing an early, clear error
    - `prepare_search_space_report` now passes `chunking_methods` and `chunk_sizes` into the payload
      and calls `prepare_search_space_with_ogx`

  ## Testing

  - Added tests for chunking overrides, individual and combined, via `prepare_search_space_with_ogx`
  - Added tests for duplicate deduplication in `chunking_methods` and `chunk_sizes`
  - Added `TestValidateChunkSizes` (11 tests) covering: `None`, valid list, single element,
    non-list, empty list, zero, negative, float, below-min, above-max, and exact boundary values
  - Added `TestPrepareSearchSpaceReportChunkSizesValidation` (3 tests) covering: non-list, empty
    list, and non-positive integer
  - Simplified `TestUnsupportedChunkingMethods` — validation fires before any I/O so complex
    mocking is no longer needed

  ## Checklist

  - [x] Tests added/updated
  - [ ] Documentation updated
  - [x] Code follows style gu


ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -17,7 +17,7 @@
 from ai4rag.core.experiment.mps import ModelsPreSelector
 from ai4rag.rag.embedding.base_model import BaseEmbeddingModel
 from ai4rag.rag.foundation_models.base_model import BaseFoundationModel
-from ai4rag.search_space.prepare.prepare_search_space import prepare_search_space_with_ogx
+from ai4rag.search_space.prepare.prepare_search_space import prepare_search_space_custom
```
filip-komarzyniec: Why do we need to change the name? 

The idea behind naming (as far as I understand @jakub-walaszczyk intention) was that there could be multiple providers (other than OGX). 
The function in its body also perform multiple validation steps strictly associated with OGX server, so the `ogx` in its name is quite significant. 
I'd say the ogx-connected logic is more important in this case than any customizability of the function.

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -268,3 +281,18 @@ def _validate_chunking_methods(methods: list[str] | None) -> None:
     for i, m in enumerate(methods):
         if not isinstance(m, str) or not m.strip():
```
filip-komarzyniec: To be honest I'm not sure this is a good pythonic code. It feels like a manual workaround to bypass Python's natural duck-typing. 
I believe we should type-hint our code and check whether it contains expected, valid values or sometimes (when justified) whether it conforms to some pre-defined protocols or parent-classes (by `issubclass`, `hasattribute` and similar checks). 

Validating everything like so, in my opinion, just makes it less python and increases overall complexity. 
@Mateusz-Switala  @jakub-walaszczyk does that resonate with you guys?

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -258,7 +269,9 @@ def _validate_model_list(models: list[str] | None, name: str) -> None:
 
 
 def _validate_chunking_methods(methods: list[str] | None) -> None:
-    """Validate that chunking methods, if provided, are non-empty strings."""
+    # TODO: remove — structural validation is now covered by AI4RAGConstraints (pydantic) and  # pylint: disable=fixme
+    #       supported-methods check is covered by prepare_search_space_custom.
```
filip-komarzyniec: What is the point of the comment? Are these functions to be removed? 

Does the pylint pragma even work if it is not on a separate line?

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -62,60 +167,113 @@ def prepare_search_space_with_ogx(
...
+
+    fms_param, ems_param = _build_model_params(foundation_models, embedding_models)
+
+    return AI4RAGSearchSpace(
+        params=[fms_param, ems_param],
+        vector_store_type=vector_store_type,
+    )
+
+
+def prepare_search_space_custom(
```
filip-komarzyniec: Is that really necessary to introduce a separate function with a little bit changed signature? 
The way I see it: 

- the possibility to introduce additional things in `payload` attribute has always been there (one could pass it to `AI4RAGSearchSpace`). The `prepare_..._with_ogx` function limited that unnecessarily. 
- now with two separate functions we have even more private functions extracting common code. So instead of 1 modified function we now ended up with >5 functions

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -62,60 +167,113 @@ def prepare_search_space_with_ogx(
...
+
+    Raises
+    ------
+    SearchSpaceValueError
+        Raised when payload contains a non-recognized parameter name or
+        when *client* is not an :class:`OgxClient`.
+    """
+    logger.info("Preparing custom search space based on provided constraints: %s.", payload)
+
+    validated_payload = AI4RAGConstraints(**payload)
```
filip-komarzyniec, line 164: How does it work if the PR does not change the `AI4RAGConstraints` model? The validation should fail here if the `payload` consists of keys other than `*_model`

ai4rag/components/optimization/search_space_preparation.py
Mateusz-Switala: But suffix `_ogx` suggest that we prepare search space using somehow ogx framework, for chunking_methods and chunk_sizes params we do not need to use ogx to validate they values. It would be misleading to put everything into one function and to keep users under a false impression that we validate ALL customised params using ogx... For models it make sense because we can check whether these models are really registered on the server.

ai4rag/components/optimization/search_space_preparation.py
Mateusz-Switala: That is why I added the comment:
```
# TODO: remove — structural validation is now covered by AI4RAGConstraints (pydantic) and  # pylint: disable=fixme
    #       supported-methods check is covered by prepare_search_space_custom.
```

I would suggest removing these two functions, which validate the chunking methods (chunking_methods) and chunk sizes (chunk_sizes), and relying solely on pydanting or some other type of validation performed at a deeper level than the component.

ai4rag/components/optimization/search_space_preparation.py
Mateusz-Switala: That was supposed to be one of the topics of the morning discussion... I am going to remove the validation methods in next commit.

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Yes, it seems to be necessary...  the question is only whether this or different name. Based on the current structure of ai4rag components and to keep backward compatibility, introducing a new function is kind of trade-off between usability and code clearance.

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Don't understand... Could you elaborate a bit? The new params have been added to AI4RAGConstraints.

ai4rag/components/optimization/search_space_preparation.py
Mateusz-Switala: The validation methods for chunking_methods and chunk_sizez have been removed!

ai4rag/components/optimization/search_space_preparation.py
Mateusz-Switala: The validation methods for chunking_methods and chunk_sizez have been removed!

ai4rag/search_space/prepare/prepare_search_space.py
filip-komarzyniec: I must have missed the changes to `AI4RAGConstraints`.. all good

tests/unit/ai4rag/components/optimization/test_search_space_prep.py
```diff
@@ -172,107 +171,3 @@ def test_non_list_models_raises_type_error(self, mock_ogx_client):
...
-        with pytest.raises(TypeError, match=r"chunking_methods\[1\] must be a non-empty string"):
-            _validate_chunking_methods(["recursive", 42])  # type: ignore[list-item]
-
-
-# ---------------------------------------------------------------------------
-# prepare_search_space_report — unsupported chunking_methods
-# ---------------------------------------------------------------------------
-
-
-class TestUnsupportedChunkingMethods:
```
filip-komarzyniec, line 242: It seems like this test makes sure that providing unsupported chunking methods raise an error. 
Why do we delete such test now?

filip-komarzyniec: oh okay, it seems that it has been moved to different file.

filip-komarzyniec (approved):
🆗 /lgtm

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -18,104 +18,210 @@
...
+    platform.  Chunking parameters (``chunking_methods``, ``chunk_sizes``) are
+    validated locally against :class:`~ai4rag.utils.constants.ChunkingConstraints`
+    and, when provided, override the platform defaults for those dimensions.
+
+    Parameters
+    ----------
+    payload : dict[str, Any]
+        A mapping of constraint names to their values.  Supported keys:
+
+        - ``"foundation_models"`` *(list[dict])* — foundation model identifiers
```
filip-komarzyniec: what exactly is this syntax of double backticks containing quotes, the `*(` the `:class:`, etc? 

These seem quite hard to replicate manually and frankly they make everything hard to read. I cannot recall such constructs nor from google- nor numpy-styled docstrings.

Do we need that?

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -18,104 +18,210 @@
...
+        Raised when payload contains a non-recognized parameter name,
+        when *client* is not an :class:`OgxClient`, when *chunking_methods*
+        contains unsupported values, or when *chunk_sizes* are out of range.
+    """
+    logger.info("Preparing search space based on provided constraints: %s.", payload)
+
+    validated_payload = AI4RAGConstraints(**payload)
+
+    if validated_payload.chunking_methods is not None:
+        unsupported = [m for m in validated_payload.chunking_methods if m not in ChunkingConstraints.METHODS]
```
filip-komarzyniec: This should not be here. The `validated_payload`, as the name suggests, should be validated. IT's either unnecessary or validation logic lacks proper validation

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -18,104 +18,210 @@
...
+        if unsupported:
+            raise SearchSpaceValueError(
```
filip-komarzyniec: The error should also be thrown during validation

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -18,104 +18,210 @@
...
+        if lang is not None:
+            fm.language = Language(**lang)
+            logger.info("Model %s: language set to %s (%s).", fm.model_id, lang["name"], lang["code"])
+        else:
+            logger.warning("Model %s: language detection failed, falling back to auto-detect.", fm.model_id)
+
 
-    fms_param = Parameter(name="foundation_model", values=foundation_models)
-    ems_param = Parameter(name="embedding_model", values=embedding_models)
+def _build_model_params(foundation_models: list, embedding_models: list) -> tuple[Parameter, Parameter]:
```
filip-komarzyniec: the whole funcion is two lines of actual code --> does it really have to be a separate function if this is called once?

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -18,104 +18,210 @@
...
+
+    extra_params: list[Parameter] = []
+    if validated_payload.chunking_methods is not None:
+        deduped_methods = list(dict.fromkeys(validated_payload.chunking_methods))
+        if len(deduped_methods) < len(validated_payload.chunking_methods):
+            logger.warning("Duplicate chunking_methods detected and removed: %s.", validated_payload.chunking_methods)
+        extra_params.append(Parameter(name=AI4RAGParamNames.CHUNKING_METHOD, values=tuple(deduped_methods)))
+    if validated_payload.chunk_sizes is not None:
+        deduped_sizes = list(dict.fromkeys(validated_payload.chunk_sizes))
+        if len(deduped_sizes) < len(validated_payload.chunk_sizes):
```
filip-komarzyniec: this is also a validation thing that should be done during validation stage

filip-komarzyniec (approved):
some comments to think through, everything else totally fine 👌🏻

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Fixed

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Done, all validations now moved to AI4RAGConstraints

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Done

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Removed the function and move the these lines back to the prepare_search_space_with_ogx

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Done

filip-komarzyniec (approved):

PR #94, opened 2026-07-07
fix: add chunk overlap param

Mateusz-Switala:
Assisted-by: Claude Code

## Description
Add chunk_overlaps param to the prepare_search_space_report component. The chunk size (chunk_size) and chunk overlap (chunk_overlap) are related by certain rules, which means that passing a custom list of chunk sizes with default overlaps can result in empty search spaces.

## Motivation
Why is this change needed?

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -114,6 +114,7 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
     random_seed: int = _DEFAULT_SEED,
     chunking_methods: list[str] | None = None,
     chunk_sizes: list[int] | None = None,
+    chunk_overlaps: list[int] | None = None,
     inference_max_threads: int = 10,
 ) -> SearchSpaceReport:
     """Run model pre-selection and prepare a search-space report.
@@ -158,6 +159,10 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         When provided, constrains the ``chunk_size`` dimension of the
         search space to only these sizes (e.g. ``[256, 512]``).  ``None``
         uses the platform defaults.
+    chunk_overlaps
+        When provided, constrains the ``chunk_overlap`` dimension of the
+        search space to only these values (e.g. ``[0, 128]``).  ``None``
+        uses the platform defaults.
     inference_max_threads
         Maximum number of concurrent threads used when querying the
         RAG service during benchmark evaluation.  Lower values reduce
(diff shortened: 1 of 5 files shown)


ai4rag/search_space/prepare/input_payload_types.py
```diff
@@ -59,6 +65,22 @@ def _deduplicate_chunk_sizes(cls, v):
             return list(dict.fromkeys(v))
         return v
 
+    @field_validator("chunk_overlaps", mode="before")
+    @classmethod
+    def _reject_bool_chunk_overlaps(cls, v):
```
filip-komarzyniec: why would someone pass boolean values as chunk_overlaps? Is that a necessary check?

ai4rag/search_space/prepare/prepare_search_space.py
```diff
@@ -171,6 +174,10 @@ def prepare_search_space_with_ogx(
         payload.get("chunk_sizes", [])
     ):
         logger.warning("Duplicate chunk_sizes removed: %s.", payload["chunk_sizes"])
+    if validated_payload.chunk_overlaps is not None and len(validated_payload.chunk_overlaps) < len(
```
filip-komarzyniec: I'm wondering whether this is a valid scenario to take into account duplicated chunks. 
Will this actually result in a broken search space? 
Introducing more code introduces more vectors for possible errors. Is being bullet-proof really give us so many benefits?

filip-komarzyniec (approved):
Looks great 🙌🏼 /lgtm

ai4rag/search_space/prepare/input_payload_types.py
Mateusz-Switala: Agree, it is an edge case, removed for chunk_sizes, chunk_overalps

ai4rag/search_space/prepare/prepare_search_space.py
Mateusz-Switala: Yeah, may not be valid... The check has been removed for chunking_methods, chunksizes and chunk_overlaps

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -246,11 +256,13 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
...
+    # Build verbose representation from valid (rule-filtered) combinations only
+    valid_combinations = search_space.combinations
+    if not valid_combinations:
+        _logger.warning("No valid combinations remain after applying search space rules.")
+    non_model_keys = [p.name for p in search_space.params if p.name not in ("foundation_model", "embedding_model")]
     verbose_repr: dict[str, Any] = {
-        k: v.all_values()
-        for k, v in search_space._search_space.items()  # pylint: disable=protected-access
-        if k not in ("foundation_model", "embedding_model")
+        key: list(dict.fromkeys(combo[key] for combo in valid_combinations)) for key in non_model_keys
```
filip-komarzyniec, line 265: I'm pretty sure the new implementation for creating the `verbose_repr` is worse in terms of time-complexity. 

I asked Claude to calculate that and here's the detailed breakdown:


Image

Whereas the previous implementation is simply `O(P)`, where `P` is the number of parameters in the search space. 

Why did we have to refactor that at all?

Mateusz-Switala: It seems that previous implementation does not work properly: 

```python
prepare_search_space_report(
            test_data_path="dummy.json",
            extracted_text_path="dummy_dir",
            ogx_client=client,
            chunking_methods=["recursive"],
            chunk_sizes=[128, 129],
        )

# {'chunking_method': ('recursive',), 'chunk_size': (128, 129), 'chunk_overlap': (0, 128, 256), 'retrieval_method': ('simple',), 'window_size': (0,), 'number_of_chunks': (3, 5, 10), 'search_mode': ('vector', 'hybrid'), 'ranker_strategy': ('', 'rrf', 'weighted'), 'ranker_k': (0, 60), 'ranker_alpha': (1, 0.5), ...}
```

Please notice, that after passing only chunk_sizes we get invalid values for chunk_overlap: ` 'chunk_size': (128, 129), 'chunk_overlap': (0, 128, 256)`  . It seems that no single rule was applied here. If this is true, then what was the purpose of returning `AI4RAGSearchSpace` in `prepare_search_space_with_ogx`? 

I agree with the analysis, but then, what is the purpose of the `search_space_preparation` component? Is this solely a “superficial” verification of the correctness of the values, with no intention of checking the relationships between parameters?

filip-komarzyniec: > it seems that no single rule was applied here.

The rules are applied dynamically during `combinations` property access. The `verbose_repr` contains `Parameter` classes' values that are the basis for building up the search space. 

As I understand that your newest changes are semantically very correct (prepare search space report does all of the preprocessing and returns a ready-to-use search space for recreation only in the subsequent component), they will not bring any values, as all of the rules will still be applied during `combinations` access.

Mateusz-Switala: Ok, understand now. I will recommend to merge it as the overall loss is marginal here. Also, I am going to open an issue to refactor AutoRAG pipelines to make it more efficient and semantically correct.

PR #101, opened 2026-07-13
feat(Timeout): Introduce efficient timeout mechanism for OGX client.

MichalSteczko:
Assisted by: Claude Code

## Description
This PR introduces efficient timeout mechanism to increase the timeout when needed during the connections with embedding and inference models.

## Motivation
Failures during longer responses from models.

## Changes
- Timeout mechanism during connection with llm model
- Timeout mechanism during connection with embedding model
- test adjustment

## Testing
By sending large requests to a models.

## Checklist
- [x] Tests added/updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/rag/foundation_models/ogx.py
+++ b/ai4rag/rag/foundation_models/ogx.py
@@ -5,9 +5,10 @@
 from typing import Annotated, Any
 
 from annotated_types import Ge, Gt, Le
-from ogx_client import OgxClient
+from ogx_client import APITimeoutError, OgxClient
 from pydantic import BaseModel
 
+from ai4rag import logger
 from ai4rag.rag.foundation_models.base_model import BaseFoundationModel, Language, MessageTyped
 from ai4rag.utils.constants import ChatGenerationConstants
 
@@ -32,7 +33,6 @@ class OGXFoundationModel(BaseFoundationModel[OgxClient, dict[str, Any] | OGXMode
         context_template_text: str | None = None,
         language: Language | None = None,
     ):
-
         super().__init__(
             client=client,
             model_id=model_id,
@@ -58,9 +58,14 @@ class OGXFoundationModel(BaseFoundationModel[OgxClient, dict[str, Any] | OGXMode
         else:
             self._params = OGXModelParameters()
 
+    _FALLBACK_TIMEOUT = 1200.0
+
     def chat(self, messages: list[MessageTyped], **kwargs) -> list[MessageTyped]:
-        """
-        Chat completion for communication with selected foundation model.
+        """Chat completion for communication with selected foundation model.
+
+        On ``APITimeoutError``, retries once with a 20-minute timeout
+        and no client-level retries to accommodate slow (CPU-deployed)
+        models.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


piotrhm (approved):
LGTM

ai4rag/rag/foundation_models/ogx.py
```diff
@@ -58,6 +58,7 @@ def params(self, params: dict | OGXModelParameters | None) -> None:
         else:
             self._params = OGXModelParameters()
 
+    @with_adaptive_timeout(max_retries=5, base_timeout=90.0, timeout_multiplier=2.0, max_timeout=1200.0)
```
Mateusz-Switala: `OgxClient` has its own retry mechanism: https://github.com/ogx-ai/ogx-client-python/blob/main/src/ogx_client/_client.py#L94 and https://github.com/ogx-ai/ogx-client-python/blob/main/src/ogx_client/_base_client.py#L751 How do your retry mechanism differ from this one already implemented by OGX?  If we merge your PR, won't we end up with a double retry in some cases?

MichalSteczko: @Mateusz-Switala It only increases the time between sending the requests, but not the time needed to get the response back. I have investigated the OGX client code base already and I Am 100% sure that my above statement is true.

Mateusz-Switala: Could you recheck that? :) It seems that your mechanism works the same way as the retry mechanism in Ogx, just at a different level.

jakub-walaszczyk: Did you manage to resolve this discussion? Do we duplicate OGX behaviour?

piotrhm: Based on our benchmarking results, timeouts are by far the most common type of error we are encountering. The solution proposed in this PR has already been partially tested and it actually improves the current situation.

jakub-walaszczyk:
## Concern: Worst-case retry overhead can reach ~21 hours for 15k documents

I ran the numbers on `with_adaptive_timeout` as configured on `OGXEmbeddingModel._call_embedding_api`:

```python
@with_adaptive_timeout(max_retries=5, base_timeout=90.0, timeout_multiplier=2.0, max_timeout=1200.0)
```

### Per-batch breakdown (5 timeouts before success on 6th attempt)

| Attempt | Timeout (waiting for failure) | Backoff delay | Subtotal |
|---------|-------------------------------|---------------|----------|
| 0 | 90s | ~91s | ~181s |
| 1 | 180s | ~181s | ~361s |
| 2 | 360s | ~361s | ~721s |
| 3 | 720s | ~721s | ~1,441s |
| 4 | 1,200s (capped) | ~1,200s (capped) | ~2,400s |
| 5 | succeeds | — | ~0s |
| **Total** | **2,550s** | **~2,554s** | **~5,104s (~85 min)** |

With `_BATCH_SIZE = 1024` and 15k documents → `ceil(15000 / 1024) = 15 batches`.

**If every batch exhausts all retries: 15 × 85 min ≈ 21.3 hours.**

### Root cause

The backoff delay formula in `ogx_retry.py:91`:
```python
delay = base_timeout * (backoff_multiplier ** attempt)
```
uses the same `base_timeout` and an identical multiplier (`2.0`) as the timeout itself. This means the "sleep between retries" is almost as expensive as the timeout wait — you spend ~50% of the total time just sleeping. By attempt 4, both the timeout and the backoff hit the `max_timeout=1200s` cap, burning **40 minutes** on a single retry cycle (20 min timeout + 20 min backoff).

### Suggestion

Consider decoupling the backoff from the timeout — e.g. a shorter fixed base for the backoff delay (say 5–10s) with its own cap, independent of `base_timeout` / `max_timeout`. The timeout should grow aggressively (as it does now) to give the server more time, but the backoff between retries doesn't need to mirror that growth — its purpose is just to avoid hammering the server immediately after a failure.

ai4rag/rag/foundation_models/ogx.py
Mateusz-Switala: Yes, currently we are duplicating ogx retry mechanism, i.e. for ogx  max_retries=3 and ai4rag max_retries=5, in the worse case, we will have 15 retries.

Mateusz-Switala (approved):
/lgtm

PR #104, opened 2026-07-16
feat: add more extenstions for files

MRGuziX:
Summary

  Expands the set of supported document formats for text extraction and discovery beyond the original six (PDF, DOCX, PPTX, Markdown, HTML, plain text).

  Changes

  - New constants.py module — extracts SUPPORTED_EXTENSIONS into a shared constants file (ai4rag/components/data/constants.py) to avoid duplication between documents_discovery.py
  and text_extraction.py
  - 9 new formats added: .odt, .odp, .adoc, .tex, .epub, .eml, .qmd, .rmd, .xhtml
  - Docling format options — registers OdtFormatOption, OdpFormatOption, AsciiDocFormatOption, LatexFormatOption, EpubFormatOption, and EmailFormatOption in the DocumentConverter
  pipeline
  - Dependency updates — upgrades docling from ~=2.107.0 to ~=2.112.0, docling-core from ~=2.84.0 to ~=2.87.1, adds odfdo>=3.0 for OpenDocument support
  - Tests updated — discovery and extraction unit tests cover all 12 supported extensions
  - Case-insensitive extension matching — uses .rmd (lowercase) in the constant set to match the .suffix.lower() check in text_extraction.py
  
  Tests
  - Tested on dev build:
  
image

image


pyproject.toml
```diff
@@ -56,6 +56,10 @@ Documentation = "https://ibm.github.io/ai4rag/"
 
 
 [project.optional-dependencies]
+feat-chunking = [
+    "docling-core[chunking]>=2.73.0,<3.0.0",
+]
+
```
filip-komarzyniec: AS discussed, we do not want to pin down both `docling` and `docling-core` as one relies on another. 

If we need `docling-core[chunking]` then we want to pin down only `docling[feat-chunking]`

ai4rag/components/data/text_extraction.py
```diff
@@ -367,6 +392,12 @@ def _build_docling_format_options(do_table_structure: bool = False) -> dict:
         InputFormat.PPTX: PowerpointFormatOption(pipeline_options=paginated_pipeline_options),
         InputFormat.HTML: HTMLFormatOption(),
         InputFormat.MD: MarkdownFormatOption(),
+        InputFormat.ODT: OdtFormatOption(),
+        InputFormat.ODP: OdpFormatOption(),
+        InputFormat.ASCIIDOC: AsciiDocFormatOption(),
+        InputFormat.LATEX: LatexFormatOption(),
+        InputFormat.EPUB: EpubFormatOption(),
+        InputFormat.EMAIL: EmailFormatOption(),
```
filip-komarzyniec, line 381: There are 9 new formats and only 6 new pipeliens. 

What about the rest of the formats? How do they aling with existing options.

filip-komarzyniec, line 381: No arguments to the new pipelines? 

Did we benchmark or run any tests on the new formats? Are they working good enough yielding sensible results?

MRGuziX: - .qmd → Docling maps it to InputFormat.MD → already handled by MarkdownFormatOption()
  - .Rmd → Docling maps it to InputFormat.MD → already handled by MarkdownFormatOption()
  - .xhtml → Docling maps it to InputFormat.HTML → already handled by HTMLFormatOption()
  
  There isnt even possible do add rest of the formats there

pyproject.toml
MRGuziX: oi

ai4rag/components/data/text_extraction.py
MRGuziX: Those formats do not have the configurable pipeline options - they are much simpler than pdf.

Benchmark - unit tests added, and also will add the functional test in autox ci repo

MRGuziX: image
@Mateusz-Switala

MRGuziX: @filip-komarzyniec 
image
Test pass with all the formats that was added

pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "pygam~=0.12.0",
     "scikit-learn==1.8.*",
     "unitxt~=1.26.1",
+    "pytest>=9.1.1",
```
jakub-walaszczyk: `pytest` should be and I believe it is an `ai4rag[test]` dependency

jakub-walaszczyk (approved):

pyproject.toml
```diff
@@ -34,8 +34,8 @@ dynamic = ["version"]
 
 dependencies = [
     "boto3>=1.28",
-    "docling~=2.107.0",
-    "docling-core[chunking]~=2.84.0",
+    "docling[feat-chunking]~=2.112.0",
+    "odfdo>=3.0",
```
filip-komarzyniec: docling ships it along `[format-opendocument]` extras

I'd say we use this in order not to cause incompatibility between ourselves-fixed `odfdo` and docling.

dev_utils/file_store.py
```diff
@@ -17,7 +17,24 @@
...
 
-SUPPORTED_EXTENSIONS = {".pdf", ".docx", ".pptx", ".md", ".html", ".txt"}
+SUPPORTED_EXTENSIONS =     {
+        ".pdf",
+        ".docx",
+        ".pptx",
+        ".md",
+        ".html",
+        ".txt",
+        # New formats
```
filip-komarzyniec: we do not want to merge such comments as they make no sense outside the scope of this PR.

filip-komarzyniec (commented):
I'd recommend updating the dependencies once again 🥶😇

pyproject.toml
filip-komarzyniec: @jakub-walaszczyk what do you think? This is after all some sort of a risk estimation. Do we want to take such gamble?

jakub-walaszczyk: If that's shipped with `docling`'s dependencies in some --extra, we should base on docling. In some time we will not know what version to use as we do not directly use this package.

Thank you @filip-komarzyniec for cautiousness. That's a great catch.

dev_utils/file_store.py
jakub-walaszczyk: @MRGuziX please remove this comment as mentioned

PR #118, opened 2026-08-03
Switch vector store to direct clients

jakub-walaszczyk:
## Description

Replace the OGX-backed vector-store layer with **direct clients** for Chroma,
Milvus, and PostgreSQL/pgvector. The backend is now selected by a typed,
frozen configuration dataclass (`ChromaConfig` / `MilvusConfig` /
`PGVectorConfig`) rather than a `vector_store_type` string plus an OGX
`vector_io` provider id. OGX remains the foundation-model and embedding
provider — only its role as a vector store is removed.

> [!WARNING]
> This is a **breaking change** to the public API (`AI4RAGExperiment`,
> `get_vector_store`, `BaseVectorStore`, `EvaluationResult`) and to generated
> pattern notebooks. See **Breaking changes** below.

## Motivation

Routing vector I/O through OGX coupled indexing/retrieval to an external
service, hid backend-specific capabilities (hybrid search, index tuning),
and made connection/secret handling implicit. Talking to each database
through its own client gives us:

- first-class **hybrid search** where the engine supports it (Milvus
  server-side dense + BM25 fusion; PGVector dense + `tsvector` full-text with
  in-memory reranking),
- explicit, typed connection configs with `from_env()` so secrets stay in the
  environment and never land in generated artefacts,
- a smaller, clearer surface (`config.provider` drives selection) and direct
  control over collection/table creation and indexing.

### Vector store core (`ai4rag/rag/vector_store/`)
- **New** `config.py` — `BaseVectorStoreConfig` + `ChromaConfig`,
  `MilvusConfig`, `PGVectorConfig` (frozen, `kw_only`, each with `from_env()`
  and `env_vars`); `get_vector_store_config()` / `get_vector_store_env_vars()`.
- **New** direct-client stores: `chroma.py`, `milvus.py`, `pgvector.py`.
- **New** `reranker.py` — `WeightedInMemoryAggregator` (RRF / weighted fusion)
  used by PGVector hybrid search.
- **Removed** `ogx.py` (`OGXVectorStore`).
- `get_vector_store(embedding_model, config, collection_name=None)` — backend
  chosen from `config.provider`; raises on config/provider mismatch.
- `base_vector_store.py` — `__init__` now takes a `config`; `collection_name`
  is a concrete property; `reuse_collection_name` removed (reuse via
  `collection_name=`). Gains `close()` (no-op default) and the context-manager
  protocol (`__enter__`/`__exit__`).
- `utils.py` — collection-name helpers (`ai4rag_<timestamp>_<suffix>`, 63-char
  cap), `validate_search_params`, embedding-dimension resolution.

### Connection lifecycle & concurrency (PGVector)
- `PGVectorStore` now opens a `psycopg_pool.ConnectionPool` (min=1, max=10)
  instead of holding a single shared `psycopg.Connection`. `search()` is
  called concurrently across threads — `query_rag` runs one worker per
  benchmark question, up to `max_threads=10` — and a single raw connection is
  not safe for that; the old manual reconnect/retry helpers are replaced by
  the pool's built-in connection health handling.
- Lazy index creation (`_ensure_indexes`) is now guarded by a lock
  (double-checked locking), since concurrent threads could otherwise all
  observe the index-built flag as unset and race on
  `CREATE INDEX IF NOT EXISTS` — PostgreSQL does not make that statement
  atomic across concurrent sessions, so the loser raised a real
  `UniqueViolation` instead of a silent no-op. A `UniqueViolation` catch
  remains as a second line of defense for a collection shared across store
  instances.
- `ChromaVectorStore` / `MilvusVectorStore` now close their underlying client
  in `close()`.

### Experiment & search space
- `AI4RAGExperiment`: removed `vector_store_type`, `ogx_vector_io_provider_id`,
  and `client`; added `vector_store_config`. Hybrid + Chroma now raises
  `RAGExperimentError`. Result payload keys `provider_type` / `collection_name`.
- `run_single_evaluation` now scopes each trial's vector store in
  `with vector_store: ...`, closing it once indexing/retrieval/generation
  finish. Previously a fresh store (and, for PGVector, a fresh DB connection)
  was created per trial and never released, leaking one connection per trial
  over a long HPO search.
- `AI4RAGSearchSpace` / `default_search_space`: `vector_store_type` default
  `"ogx"` → `"milvus"`; supported values `"milvus"`, `"pgvector"`, `"chroma"`
  (Chroma excludes hybrid params).

### Assets generator
- Removed `pattern_builder.py` (`build_pattern_json`); template placeholders
  updated (`PROVIDER_TYPE`, `COLLECTION_NAME`, `REQUIRED_ENV_VARS`).
- `run_rag_optimization` takes `vector_store_config` instead of
  `vector_io_provider_id` (still accepts `ogx_client` for models).

### Dependencies (`pyproject.toml`)
- Added `chromadb>=1.5,<2`, `pymilvus>=2.5.0`, `pgvector>=0.3.0`,
  `psycopg[binary,pool]>=3.1` (the `pool` extra backs `PGVectorStore`'s
  connection pool); removed `langchain_chroma`; kept `ogx-client`
  (models/embeddings). Pylint `ignored-modules = ["psycopg"]`.

### Documentation
- Rewrote the vector-stores API page (mkdocstrings for `config`,
  `get_vector_store`, the three stores, `reranker`, `utils`); dropped the
  deleted `vector_store.ogx` and `pattern_builder` blocks.
- Reframed OGX as the model/embedding provider across architecture,
  user-guide, getting-started, `index.md`, and `README.md`; migrated every
  `AI4RAGExperiment` example to `vector_store_config`; corrected the
  hybrid-search support matrix; fixed the data-flow retrieval diagram (the
  store embeds the raw query). Verified with `mkdocs build --strict`.

## Breaking changes

- `AI4RAGExperiment(...)`: `vector_store_type`, `ogx_vector_io_provider_id`,
  and `client` removed — pass `vector_store_config=<Chroma|Milvus|PGVector>Config`.
- `get_vector_store(...)`: new `(embedding_model, config, collection_name=None)`
  signature.
- `BaseVectorStore`: `config`-based `__init__`; `reuse_collection_name` removed.
- `OGXVectorStore` and `build_pattern_json` removed.
- Search space no longer accepts `vector_store_type="ogx"`.
- `EvaluationResult` no longer carries a `rag_pattern` field. A trial's vector
  store is now closed once the trial finishes, so keeping a reference to its
  RAG pattern would leave callers holding a template whose retriever points at
  a closed store. Read `pattern_name` / `scores` from `EvaluationResult`
  instead of calling `.generate()` on a previously returned pattern.

## Testing

- **Unit** — new suites for `config`, `milvus`, `pgvector`, `reranker`, and
  expanded `chroma` / `get_vector_store` / `utils`; removed `test_ogx.py` and
  `test_pattern_builder.py`; updated experiment / optimization / templates.
  Added a regression test for `PGVectorStore._ensure_indexes` that runs 8
  threads concurrently against a latency-injected mock connection to
  reproduce and guard against the index-creation race (verified it reliably
  fails without the fix and reliably passes with it).
- **Integration** — DB-only suites with deterministic embeddings for Chroma,
  Milvus, PGVector (`tests/integration/vector_store/`).
- **Functional** — real-model suites for each backend
  (`tests/functional/vector_store/`).
- **Docs** — `mkdocs build --strict` passes (exit 0).

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [ ] All checks passing


jakub-walaszczyk:
fyi @filip-komarzyniec @MichalSteczko

ai4rag/rag/vector_store/config.py
```diff
@@ -0,0 +1,314 @@
...
+            port=int(os.environ.get("PGVECTOR_PORT", "5432")),
+            dbname=os.environ.get("PGVECTOR_DB", "postgres"),
+            user=os.environ.get("PGVECTOR_USER", "postgres"),
+            password=os.environ.get("PGVECTOR_PASSWORD"),
+        )
+
+
+# Registry mapping a provider discriminator to its config class. Built from each
+# class's ``provider`` default so the provider string has a single source of truth.
+_CONFIG_BY_PROVIDER: dict[str, type[BaseVectorStoreConfig]] = {
```
MichalSteczko: While using dict as GLOBAL variable please use `NamedTuple` or `types.MappingProxyType`. In Python 3.15 it will be entirely replaced with builtin `frozendict`. We shouldn't leave it mutable.

MichalSteczko (commented):
Overall really good job was made here!
General comment for the "concurrency" please do not use "up to 10" concurrent connection if there are less than 10 threads, it will significantly slow down the entire operation instead of speeding it up. In any other "connection" related stuff (e.g vector store connections) please use async operations it will also remove the "lock" (race conditions) overhead.

ai4rag/rag/vector_store/config.py
jakub-walaszczyk: This one was fixed. For the concurrency, let's revisit if there are some issues. The current implementation already handles some issues that occurred during delivery

MichalSteczko: Thank you! We can work together during revisiting our concurrency model.

MichalSteczko (approved):

jakub-walaszczyk:
@MichalSteczko I have added last fix for chroma `EphemeralClient` after agentic review. All tests passing.

PR #122, opened 2026-08-06
feat: add preparation space check step to fail fast

MRGuziX:
## Description
We should fail fast when it come to the checking of space_preparation step, right now user can launch experiment that will take hours, and after that we will show him info "yo, your model is wrong" he cannot re-trigger it with a changed config, he cannot do anything to save hours of computing. So we need to fail fast to give him info that his provided input is "ok" or "nook"
## Motivation
https://redhat.atlassian.net/browse/RHOAIENG-57709

## Changes
1. ai4rag/utils/validators.py — added validate_model_list() function (moved from search_space_preparation.py)
  2. ai4rag/components/optimization/search_space_preparation.py —
    - Imports validate_model_list from shared location + AI4RAGSearchSpace
    - Adds _validate_model_list = validate_model_list alias for backward compat
    - Adds pre_validated_search_space: AI4RAGSearchSpace | None = None parameter to prepare_search_space_report()
    - When provided: skips model validation, payload construction, and prepare_search_space_with_ogx() call
    - When None: existing behavior unchanged
  3. tests/unit/ai4rag/components/optimization/test_search_space_prep.py —
    - Added TestValidateModelListShared (verifies alias identity)
    - Added TestPrepareSearchSpaceReportPreValidated with 3 tests: skips OGX call, uses provided search space, still loads documents
    

## Testing
- created dev image
- pushed it to quay
- use url in pipeline.yaml
- run a test with changed image
- Step passed - failed it is that there is ai4rag 0.11 and PLC do not have that image yet so it failed last step
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -275,5 +275,3 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         search_space=verbose_repr,
         selected_models=selected_models,
     )
-
-
--- a/ai4rag/components/optimization/search_space_preparation.py
+++ b/ai4rag/components/optimization/search_space_preparation.py
@@ -18,9 +18,13 @@ from ai4rag.core.experiment.mps import ModelsPreSelector
 from ai4rag.rag.embedding.base_model import BaseEmbeddingModel
 from ai4rag.rag.foundation_models.base_model import BaseFoundationModel
 from ai4rag.search_space.prepare.prepare_search_space import prepare_search_space_with_ogx
+from ai4rag.search_space.src.search_space import AI4RAGSearchSpace
+from ai4rag.utils.validators import validate_model_list
 
 _logger = logging.getLogger("search-space-preparation")
 _logger.addHandler(handler)
+
+_validate_model_list = validate_model_list
 
 _DEFAULT_METRIC = "faithfulness"
 _DEFAULT_TOP_N_GENERATION = 3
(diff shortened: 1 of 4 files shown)


ai4rag/utils/validators.py
```diff
@@ -66,3 +66,14 @@ def validate(self, obj, value: T):
                 f"on attribute {self.private_name[1:]}."
             )
         return value
+
+
+def validate_model_list(models: list[str] | None, name: str) -> None:
```
Mateusz-Switala, line 71: I am not sure if this is the right place for this function. Maybe, let's leave it in ai4rag/components/optimization/search_space_preparation.py

ai4rag/components/optimization/search_space_preparation.py
```diff
@@ -161,6 +166,13 @@ def prepare_search_space_report(  # pylint: disable=too-many-locals,too-many-arg
         RAG service during benchmark evaluation.  Lower values reduce
         per-request concurrency (useful when each request carries more
         retrieved context).  Defaults to ``10``.
+    pre_validated_search_space
+        When provided, the function skips model-list validation,
+        payload construction, and the
+        :func:`prepare_search_space_with_ogx` call and uses this
```
Mateusz-Switala, line 172:
```suggestion
        `prepare_search_space_with_ogx` call and uses this
```

The markup `:func:prepare_search_space_with_ogx` is Sphinx reStructuredText (rst) syntax. Let's use sth closer to google style.

ai4rag/utils/validators.py
Mateusz-Switala: @MRGuziX  Could you address this comment? Thanks

jakub-walaszczyk:
@MRGuziX what about this PR? Is this valid?

PR #126, opened 2026-08-10
feat(evaluator): add optional RAGAS evaluator

piotrhm:
## Description
Adds RAGAS as an optional, independent evaluator alongside the in-house LLM judge, exposing four RAG metrics (faithfulness, answer relevancy, context precision, context recall) that run through the pipeline's already-configured foundation and embedding models.

## Motivation
The project only offered its in-house LLM-as-a-Judge metrics. RAGAS is a widely-used, standardized RAG evaluation library, and supporting it gives users an independent, comparable set of metrics without having to stand up a separate evaluation stack. Wrapping our own model abstractions means RAGAS reuses the endpoint the rest of the pipeline is already configured with instead of opening its own OpenAI/LangChain client.

## Changes
* **New `RagasEvaluator`** `ai4rag/evaluator/ragas_evaluator.py` - scores RAG metrics via the RAGAS library and returns them in the shared `EvaluationMetricsResult` shape, with bootstrap confidence intervals computed the same way as the other evaluators. Failed/slow samples yield NaN → None instead of aborting the whole pattern evaluation.
* **RAGAS adapters** `ai4rag/evaluator/ragas_adapters.py` thin `BaseRagasLLM` / `BaseRagasEmbeddings` wrappers delegating to `BaseFoundationModel.chat` and `BaseEmbeddingModel`. All ragas/langchain imports are local so importing `ai4rag.evaluator` never requires the optional extra.
* **New metrics** `ai4rag/evaluator/metric.py` - `RAGAS_FAITHFULNESS`, `RAGAS_ANSWER_RELEVANCY`, `RAGAS_CONTEXT_PRECISION`, `RAGAS_CONTEXT_RECALL`, plus a `ragas` evaluator literal.
* **Wiring** — `run_rag_optimization` gains a `ragas_enabled: bool = False` flag that appends a `RagasEvaluator`; `AI4RAGExperiment` adds the four RAGAS metrics to its default set when a ragas evaluator is present; `RagasEvaluator` exported from `ai4rag.evaluator`.
* **Dependencies** — new optional ragas extra: ragas>=0.3,<0.5 with langchain-community<0.4.2 (pinned because ragas 0.4.3 imports a Vertex AI module langchain-community 0.4.2 removed).
* **Tuned run config** — `RunConfig(timeout=1200, max_workers=4)` to avoid overwhelming small evaluating models and to prevent spurious thread-pool `TimeoutErrors`.
## Testing
Added `tests/unit/ai4rag/evaluator/test_ragas_evaluator.py` covering the evaluator.


piotrhm:
Related `pipelines-components` PR -> https://github.com/opendatahub-io/pipelines-components/pull/204

piotrhm:
# RAG Evaluation Metrics: Legacy vs. RAGAS

## The experiment

We ran the `documents-rag-optimization-pipeline` across **7 benchmark datasets**
spanning three corpora families and multiple formats/sizes:

| Corpus | Datasets |
|---|---|
| Enterprise RAG-Bench | github-50, confluence-50 |
| Open RAG-Bench (arXiv) | arxiv-10-txt, arxiv-100-txt, arxiv-100-md |
| HTML-RAG NQ | nq-val-10, nq-val-50 |

For each dataset the pipeline performed HPO over the RAG design space (chunking, embeddings,
retrieval, generation), producing **10 candidate patterns per dataset = 70 configurations total**.
Every configuration was scored by **two independent metric families**:

- **Legacy metrics** (current selection criteria): faithfulness, answer correctness,
  context correctness, overall score.
- **RAGAS metrics** (proposed addition): faithfulness, answer relevancy, context precision,
  context recall.

We then measured how well the two families agree — per dataset (Spearman rank correlation,
n=10 each) and pooled (n=70) — to determine whether RAGAS is **redundant** (measures the same
thing) or **complementary** (adds new signal).

---

## Summary

**The legacy and RAGAS metric families do not agree, and where they diverge the legacy metrics
are the less trustworthy of the two. Adding RAGAS is warranted — it supplies discriminating
signal precisely where the current metrics fail.**

Three findings drive the recommendation:

1. **Low, unstable correlation → the metrics are complementary, not redundant.**
   Same-named pairs (e.g. legacy faithfulness vs. RAGAS faithfulness) correlate only
   **+0.37 pooled**, and swing from **+0.85 to −0.18** across datasets. If RAGAS merely restated
   the existing metrics, correlations would be uniformly high. They are not — RAGAS captures
   information the legacy metrics miss.

2. **The legacy "context correctness" metric is unreliable.** It is **saturated at a constant
   1.0** on arxiv-10-txt (zero variance — it cannot rank anything) and near-ceiling on the other
   open/NQ datasets. On the enterprise data where it does vary, it **inversely correlates** with
   RAGAS context precision on github-50 (**ρ = −0.84**). A metric that is either flat or backwards
   cannot be a sound optimization target. RAGAS context precision/recall provide the variance
   needed to actually discriminate retrieval quality.

3. **The choice of metric family changes which RAG configuration "wins"** on most datasets.
   Because the families rank patterns differently, optimizing on legacy scores alone risks
   selecting a suboptimal RAG design. A second, independent measurement reduces that risk.

---

## Best configuration per dataset (actual scores)

The best pattern per dataset (selected by legacy `overall` score), with both metric
families side by side. Legacy `context correctness` (`L.ctxCorr`) is pinned at **1.000**
on every non-enterprise dataset — a visible symptom of the saturation problem.

| dataset | best pattern | L.overall | L.faith | L.ansCorr | L.ctxCorr | R.faith | R.ansRel | R.ctxPrec | R.ctxRecall |
|---|---|---|---|---|---|---|---|---|---|
| arxiv-10-txt | Pattern7 | 0.839 | 0.711 | 0.462 | 1.000 | 0.850 | 0.931 | 0.915 | 1.000 |
| arxiv-100-txt | Pattern9 | 0.835 | 0.751 | 0.633 | 1.000 | 0.755 | 0.870 | 0.841 | 0.994 |
| arxiv-100-md | Pattern9 | 0.828 | 0.718 | 0.641 | 1.000 | 0.753 | 0.866 | 0.850 | 0.971 |
| nq-val-10 | Pattern1 | 0.782 | 0.715 | 0.465 | 1.000 | 0.867 | 0.863 | 0.675 | 0.891 |
| nq-val-50 | Pattern3 | 0.825 | 0.668 | 0.781 | 1.000 | 0.883 | 0.703 | 0.767 | 0.975 |
| confluence-50 | Pattern7 | 0.742 | 0.570 | 0.481 | 0.828 | 0.764 | 0.719 | 0.854 | 0.979 |
| github-50 | Pattern1 | 0.690 | 0.533 | 0.433 | 0.704 | 0.684 | 0.626 | 0.896 | 0.951 |

_`L.` = legacy metric, `R.` = RAGAS metric. All winning configs share `bge-m3` embeddings,
`redhataillama-31-8b-instruct` generation, and `recursive` chunking._

best_scores_per_dataset


> Legacy overall vs. RAGAS faithfulness for the best pattern of each dataset. On arxiv-100-txt/md
> the two families disagree by ~7–8 points in opposite directions; the apparent "winner" depends on
> which family you trust.

---

## Supporting data

### Agreement between "matched" metric pairs (Spearman ρ, n=10 per dataset)

correlation_heatmap

| dataset | faith ↔ ragas_faith | ctx_corr ↔ ragas_ctx_prec | overall ↔ ragas_faith |
|---|---|---|---|
| arxiv-10-txt | +0.25 | **n/a (flat)** | +0.50 |
| arxiv-100-txt | −0.04 | +0.75 | +0.67 |
| arxiv-100-md | +0.49 | +0.57 | +0.78 |
| nq-val-10 | −0.18 | +0.70 | +0.13 |
| nq-val-50 | +0.85 | +0.79 | +0.89 |
| confluence-50 | +0.70 | +0.41 | +0.71 |
| github-50 | +0.22 | **−0.84** | +0.49 |
| **Pooled (n=70)** | **+0.37** | **+0.16** | **+0.56** |

> If the two families were interchangeable, every cell would be ≈ +0.9. Instead they range from
> strongly positive to negative — evidence of complementary signal, not duplication.

### The legacy context metric has no discriminating power on most datasets

ctx_corr_saturation

> Legacy context correctness across all 10 patterns per dataset. It sits at the 1.0 ceiling on the
> open/NQ datasets (no ability to rank) and only spreads out on confluence-50 and github-50.

| dataset | distinct legacy `context_correctness` values (10 patterns) |
|---|---|
| arxiv-10-txt | `[1.0]` — **constant, cannot rank** |
| arxiv-100-txt | `[0.92, 0.96, 1.0]` — ceiling-bound |
| arxiv-100-md | `[0.94, 0.98, 1.0]` — ceiling-bound |
| nq-val-10 | `[0.9, 1.0]` — near-binary |
| confluence-50 | `[0.70 … 0.85]` — healthy spread |
| github-50 | `[0.68 … 0.77]` — healthy spread |

> The legacy context metric only varies meaningfully on the two enterprise datasets — and that is
> exactly where it disagrees with RAGAS (github ρ = −0.84). Everywhere else it is saturated and blind.

### Most reliable cross-family signal

`overall ↔ RAGAS faithfulness` is **positive on all 7 datasets** (+0.13 to +0.89, pooled +0.56) —
RAGAS faithfulness is the single best independent corroborator of the current aggregate score,
making it a natural first RAGAS metric to adopt.

---

## Recommendation

Adopt RAGAS as a **parallel, complementary** evaluation family — not a replacement. Concretely:

- **Add RAGAS faithfulness + answer relevancy** as reported metrics immediately (relevancy is the
  strongest driver of the current `overall`, ρ up to +0.87).
- **Replace or supplement legacy "context correctness" with RAGAS context precision/recall**, given
  the legacy metric's ceiling saturation and sign inversion.
- **Flag configurations where the two families disagree** for human review before promoting a
  "best" RAG pattern.

---

_Basis: 7 datasets × 10 patterns = 70 configurations, from the latest 10-pattern
`documents-rag-optimization-pipeline` run per dataset (enterprise runs completed 2026-08-10).
Correlations are Spearman rank, computed within-dataset to remove cross-corpus difficulty confounds._

.github/workflows/pr-checks.yml
```diff
@@ -21,7 +21,9 @@ jobs:
           python-version: '3.13'
 
       - name: Install dependencies
-        run: uv sync --extra code_check
+        # Include the optional ``ragas`` extra so pylint can resolve the RAGAS
+        # evaluator's lazy imports instead of reporting import-error.
+        run: uv sync --extra code_check --extra ragas
```
jakub-walaszczyk: We can add ragas as normal deps, no extras. They will be always used in the projects consuming ai4rag.

ai4rag/components/optimization/rag_templates_optimization.py
```diff
@@ -75,6 +76,7 @@ def run_rag_optimization(  # pylint: disable=too-many-locals,too-many-arguments,
     inference_max_threads: int = 10,
     indexing_pipeline_params: dict | None = None,
     judge_enabled: bool = True,
+    ragas_enabled: bool = False,
```
jakub-walaszczyk: If judge is enabled, let's use that condition to handle both evaluators, as ragas strongly uses llmaaj ideas. Seems to be fair

ai4rag/evaluator/base_evaluator.py
```diff
@@ -101,6 +103,43 @@ class EvaluationMetricsResult(TypedDict):
     question_scores: list[QuestionScore]
 
 
+def build_aggregate_metric(
```
jakub-walaszczyk, line 106: fair idea, but unitxt_evaluator provides his own implementation in different way. Can it be somehow reused so that there is no unitxt_evaluator own implementation overhead?

ai4rag/evaluator/metric.py
```diff
@@ -42,6 +42,27 @@ class Metrics(metaclass=ConstantMeta):
         description="LLM judge score for how directly and helpfully the response addresses the question.",
     )
 
+    RAGAS_FAITHFULNESS = RAGMetric(
+        name="ragas_faithfulness",
```
jakub-walaszczyk: I am wondering if this is a good idea for metric name to include its evaluator. This is why we introduced evaluator field: if metric name is duplicated we know it's from the different evaluator.

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,125 @@
...
+interfaces.  These thin wrappers delegate to the ai4rag model abstractions
+(:class:`BaseFoundationModel.chat` and :class:`BaseEmbeddingModel`), so RAGAS
+reuses whatever endpoint the rest of the pipeline is already configured with
+instead of opening its own OpenAI/LangChain client.
+
+All ``ragas`` / ``langchain`` imports are local to this module so importing
+:mod:`ai4rag.evaluator` never requires the optional ``ragas`` extra.
+"""
+
+from __future__ import annotations
```
jakub-walaszczyk: Not needed

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,125 @@
...
+
+import asyncio
+from typing import TYPE_CHECKING, Any
+
+from ai4rag.rag.embedding.base_model import BaseEmbeddingModel
+from ai4rag.rag.foundation_models.base_model import BaseFoundationModel
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
```
jakub-walaszczyk: To be removed

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,125 @@
...
+instead of opening its own OpenAI/LangChain client.
+
+All ``ragas`` / ``langchain`` imports are local to this module so importing
+:mod:`ai4rag.evaluator` never requires the optional ``ragas`` extra.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from typing import TYPE_CHECKING, Any
```
jakub-walaszczyk: type checking to be removed

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,125 @@
...
+
+def make_ragas_llm(model: BaseFoundationModel, max_completion_tokens: int = DEFAULT_MAX_COMPLETION_TOKENS) -> Any:
+    """Build a ``BaseRagasLLM`` backed by an ai4rag foundation model.
+
+    Defined as a factory so the ``ragas`` base class is imported lazily.
+    """
+    from langchain_core.outputs import Generation, LLMResult
+    from ragas.llms.base import BaseRagasLLM
+
+    class Ai4ragRagasLLM(BaseRagasLLM):
```
jakub-walaszczyk: nice and fair implementation, but why is it inside the `make_ragas_llm()` function?

Also `AI4RAGR...` I think should remain capitalized (or maybe not)

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,125 @@
...
+    from langchain_core.prompt_values import PromptValue
+
+# RAGAS metric prompts emit structured JSON (statement lists, verdicts).  The
+# ai4rag model's default ``max_completion_tokens`` can be small enough to
+# truncate that JSON, which RAGAS then fails to parse -> NaN scores.  Cap the
+# completion generously so evaluation prompts are not cut off.
+DEFAULT_MAX_COMPLETION_TOKENS = 1024
+
+
+def make_ragas_llm(model: BaseFoundationModel, max_completion_tokens: int = DEFAULT_MAX_COMPLETION_TOKENS) -> Any:
```
jakub-walaszczyk: Docstrings in this module are not full. Missing parameters docstring

pyproject.toml
```diff
@@ -60,6 +60,11 @@ dev = [
   "ipykernel",
 ]
 
+# ragas 0.4.3 imports langchain_community.chat_models.vertexai unconditionally, which
+# langchain-community 0.4.2 removed. Pin <0.4.2 (0.4.1 still ships it) until ragas drops
+# the hard Vertex AI import.
+ragas = ["ragas>=0.3,<0.5", "langchain-community<0.4.2"]
```
jakub-walaszczyk: Can we ship it to the regular dependencies? I think we will use it always in the actual AutoRAG pipeline

jakub-walaszczyk (commented):
Overall very strong PR. There are some suggestions, that seem to address LLM-generated content. Once we address them we may ship it with coming v0.12.0. 

2 additional comments:
- you should also introduce documentation update
- make sure docstrings are in place even for private methods (this is for developers in the first place)

.github/workflows/pr-checks.yml
piotrhm: done.

ai4rag/components/optimization/rag_templates_optimization.py
piotrhm: added LLMJudgeMode = Literal["base", "ragas", "all", "none"], so we have better control over selected judge method group

ai4rag/evaluator/base_evaluator.py
piotrhm: I tried to reuse it, I extracted build_aggregate_metric into base_evaluator.py, and all three evaluators (unitxt, ragas, llmaj) call it, so the mean/CI → AggregateMetric assembly is identical everywhere

ai4rag/evaluator/ragas_adapters.py
piotrhm: Removed.

ai4rag/evaluator/ragas_adapters.py
piotrhm: Removed.

ai4rag/evaluator/ragas_adapters.py
piotrhm: Simplified and renamed.

pyproject.toml
piotrhm: Done.

ai4rag/evaluator/ragas_adapters.py
piotrhm: Added.

ai4rag/evaluator/metric.py
piotrhm: Tricky one, I remove "ragas_" prefix, at the cost od introducing changes in ai4rag/core/experiment/experiment.py

ai4rag/evaluator/ragas_adapters.py
piotrhm: Done.

ai4rag/core/experiment/experiment.py
```diff
@@ -253,7 +253,13 @@ def metrics(self, val: Sequence[RAGMetric | str] | None) -> None:
...
             return
 
-        available = {m.name: m for m in Metrics}
+        # A metric name is not unique across evaluators (e.g. both the unitxt and
+        # RAGAS evaluators expose "faithfulness"); resolving a bare string keeps
+        # the first-defined metric so behaviour stays stable and back-compatible.
+        # Pass a RAGMetric instance to select a specific evaluator's variant.
+        available: dict[str, RAGMetric] = {}
+        for m in Metrics:
+            available.setdefault(m.name, m)
```
jakub-walaszczyk: I think that we should consider expecting user to provide broader information.  
  
As of today we return list of metrics, where each metric is a JSON object containing both name and evaluator. This pair makes each metric unique. In this context here we do not do that. Maybe we should expect user to provide here list actual RAGMetric instances selected from `Metric`. I would rather go for that here, and raise error when string is provided.  
  
What's more we need to make some statements on how the metrics are resolved from the user metrics in pipelines. Let's make some strong assumption at this stage, that we **use only unitxt metrics as the optimization target**. Here we will expect only valid RAGMetric instances from Metrics.  
  
Following that we will select optimization metric to be one of unitxt metrics or overall score. In the future (near perhaps) we will enable user to select llmaaj metric as optimization metric, but it will require enhancing the input parameters (work with pipelines-components, UI and docs).  
  
What do you tink?

ai4rag/evaluator/unitxt_evaluator.py
```diff
@@ -31,6 +33,15 @@ class UnitxtEvaluator(BaseEvaluator):
...
     }
 
+    # Reference-based unitxt metrics compute a token overlap against a list of
+    # references; unitxt crashes (``max()`` on an empty sequence) when that list
+    # is empty for a record. We therefore evaluate each such metric only over the
+    # records that actually carry its references — the field named below.
+    _REFERENCE_FIELD: dict[str, str] = {
+        "metrics.rag.external_rag.faithfulness": "contexts",
+        "metrics.rag.external_rag.answer_correctness": "ground_truths",
+    }
```
jakub-walaszczyk: I can see this is your response to the issue that appeared. The issue will not be resolved by this implementation for milvus vector database, as the underlying issue was actually consistency issues (adding documents and fetching documents quickly after ended up in returning empty documents, that's what was missing). You can safely restore the previous implementation. I will merge PR with fixes shortly.

jakub-walaszczyk: Or is it confirmed, that these fields make difference? I do not recall that it impacts the metrics actually.

ai4rag/evaluator/metric.py
jakub-walaszczyk: That's expected. Thanks!

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,268 @@
...
+
+        Returns
+        -------
+        list[list[float]]
+            One embedding per input document.
+        """
+        return await asyncio.to_thread(self._model.embed_documents, texts)
+
+
+def make_ragas_llm(model: BaseFoundationModel, max_completion_tokens: int = DEFAULT_MAX_COMPLETION_TOKENS) -> Any:
```
jakub-walaszczyk: Isn't it sufficient to call models constructors directly? Do we need to call a separate function just to call an underlying class's constructor?

ai4rag/evaluator/ragas_adapters.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Returns
+    -------
+    AI4RAGRagasLLM
+        A RAGAS-compatible LLM delegating to ``model``.
+    """
+    return AI4RAGRagasLLM(model, max_completion_tokens=max_completion_tokens)
+
+
+def make_ragas_embeddings(embedding_model: BaseEmbeddingModel) -> Any:
```
jakub-walaszczyk: Same as for `make_ragas_llm`

jakub-walaszczyk (commented):
Minor comments to resolve and we are getting closer and closer to the merge. The solution for milvus returning empty contexts will be merged soon 👀

ai4rag/evaluator/ragas_adapters.py
piotrhm: Good catch, you're right, they added nothing over the constructors. Removed.

ai4rag/evaluator/ragas_adapters.py
piotrhm: Removed.

ai4rag/evaluator/unitxt_evaluator.py
piotrhm: Reverted back to the single evaluate() call.

ai4rag/core/experiment/experiment.py
piotrhm: Done as suggested. Both the metrics and optimization_metric setters now expect RAGMetric instances from Metrics and raise on bare strings (ambiguous, since e.g. faithfulness exists for both unitxt and RAGAS).

jakub-walaszczyk (approved):

PR #127, opened 2026-08-12
refactor(maas)!: simplify integration to a single OpenAI client

jakub-walaszczyk:
MaaS now serves models.list(), chat.completions and embeddings from one OpenAI-compatible endpoint, so the previous one-client-per-model design (per-model URL derivation and id stripping) was dead weight.

- Drop per-model client/URL machinery (create_maas_model_client, maas_model_base_url, _short_model_id, _model_owned_by, _resolve_base_url); a single client backs list, chat and embeddings.
- Use model ids verbatim, exactly as models.list() reports them (including '/' path segments); stop stripping to the last segment.
- Move model-instantiation helpers from components/utils/models.py to ai4rag/search_space/prepare/models.py, co-located with their primary consumer, removing the components<->search_space coupling; every consumer now imports them directly.
- Drop base_url from the search-space report and pattern.json; notebooks and dev utils read MAAS_BASE_URL verbatim and build one OpenAI client for both foundation and embedding models.
- Pass the pattern's detected language into the generated inference notebook's foundation model.
- Read benchmark JSON with an explicit utf-8 encoding.
- Update docs and tests to the single-client model; remove tests that only asserted the per-model behaviour.

BREAKING CHANGE: the ai4rag.components.utils.models module and the create_maas_model_client / maas_model_base_url helpers are removed, model ids are now used verbatim (full ids), and the search-space report and pattern.json no longer carry a per-model base_url.

Assisted-by: Claude Opus 4.8

## Description
Brief summary of changes

## Motivation
Why is this change needed?

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing


MichalSteczko (approved):
LGTM

PR #128, opened 2026-08-18
feat: ai4rag ocr

Medokins:
## Description
Adds optional OCR support to the text-extraction component so scanned PDFs and
  image files (JPEG/PNG/TIFF) can be ingested. OCR runs through RapidOCR via
  Docling and is **off by default** — born-digital documents are unaffected.
  Conversion behaviour is driven by a single `DoclingExtractionConfig` passed to
  `extract_text`.
  
  ## Changes
  - **Image ingestion** — added `.jpg`, `.jpeg`, `.png`, `.tif`, `.tiff` to
    `SUPPORTED_EXTENSIONS`; images are routed through Docling's image/PDF pipeline
    so OCR can run when enabled.
  - **`DoclingExtractionConfig`** — a single frozen config object controls
    table-structure and OCR behaviour (`do_ocr`, `ocr_lang`, and optional custom
    `ocr_*_model_path` overrides). `extract_text` now accepts a ready
    `docling_config` instance and forwards it unchanged to every worker process,
    instead of exploding and re-packing individual parameters. Exported from
    `ai4rag.components.data`.
  - **Model resolution** — `_build_rapidocr_options` resolves ONNX models in a
    three-tier order: (1) explicit custom paths, (2) ONNX bundled in the installed
    `rapidocr` wheel, (3) fall back to `DOCLING_ARTIFACTS_PATH/RapidOcr` (baked
    image models).
   - **Language handling** — no automatic detection; `ocr_lang` selects the bundled
    model set (Latin/English default vs Chinese). `ocr_lang` accepts a string or a
    sequence and is normalized centrally in `DoclingExtractionConfig.__post_init__`.
  - **Fail-fast, actionable errors** —
    - `_validate_rapidocr_artifacts` raises a clear "bake these exact files" error
      when `DOCLING_ARTIFACTS_PATH` is set but the OCR ONNX models are missing
      (Docling itself only logs a warning and then fails opaquely inside a worker).
    - A non-importable `rapidocr` on the OCR path now raises an actionable
      `RuntimeError` instead of silently degrading (rapidocr is a hard dependency
      of `docling-slim[standard]`, so a failed import means a broken install).

  ## Testing
  - `pytest tests/unit` — full unit suite green (1307 passed) after `uv sync --extra dev`.
  - Extraction-specific tests (`tests/unit/ai4rag/components/data/test_extraction.py`)
    cover: image format registration, OCR enable/disable, English vs Chinese model
    selection, custom model paths, `ocr_lang` normalization, missing-artifacts
    fail-fast, and missing-`rapidocr`-package fail-fast.
  - `pylint` / `black` / `isort` clean on the changed files.


ai4rag/components/data/constants.py
```diff
@@ -19,5 +19,11 @@
         ".qmd",
         ".rmd",
         ".xhtml",
+        # Image formats for OCR-capable ingestion (RapidOCR via Docling).
```
jakub-walaszczyk: Please remove this unnecessary comment

ai4rag/components/data/text_extraction.py
```diff
@@ -75,6 +94,12 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
     max_extraction_workers: int | None = None,
     docling_artifacts_path: str | None = None,
     do_table_structure: bool = False,
+    do_ocr: bool = False,
+    ocr_lang: Sequence[str] | str | None = None,
+    ocr_det_model_path: str | None = None,
+    ocr_cls_model_path: str | None = None,
+    ocr_rec_model_path: str | None = None,
+    ocr_rec_keys_path: str | None = None,
```
jakub-walaszczyk: Don't we want to expect instance of `DoclingExtractionConfig` as a parameter here? `pipelines-components` might construct this and pass prepared instance

ai4rag/components/data/text_extraction.py
```diff
@@ -63,7 +69,20 @@ class ExtractionResult:
     error_count: int
 
 
-def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many-positional-arguments
+@dataclass(frozen=True)
+class DoclingExtractionConfig:
```
jakub-walaszczyk, line 106: This would be nice to have docstring here describing each parameter "what" and "why".

ai4rag/components/data/text_extraction.py
```diff
@@ -140,8 +181,23 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
 
     s3_creds = _resolve_s3_credentials(s3_endpoint, s3_access_key, s3_secret_key, s3_region)
     artifacts_path = _resolve_artifacts_path(docling_artifacts_path)
+    pipeline_config = DoclingExtractionConfig(
```
jakub-walaszczyk: as suggested in the class signature comment, double-handling these parameters might not have much rationale. We  can pass ready instance of this dataclass to the function while calling it.

ai4rag/components/data/text_extraction.py
```diff
@@ -350,24 +406,148 @@ def _resolve_artifacts_path(explicit: str | None) -> Path | None:
...
+)
+_ARTIFACTS_RAPIDOCR_CHINESE = (
+    "onnx/PP-OCRv4/det/ch_PP-OCRv4_det_mobile.onnx",
+    "onnx/PP-OCRv4/cls/ch_ppocr_mobile_v2.0_cls_mobile.onnx",
+    "onnx/PP-OCRv4/rec/ch_PP-OCRv4_rec_mobile.onnx",
+)
+# Optional small models that some rapidocr wheels still ship under ``rapidocr/models/``.
+_BUNDLED_RAPIDOCR_DET = "PP-OCRv6_det_small.onnx"
+_BUNDLED_RAPIDOCR_CLS = "ch_ppocr_mobile_v2.0_cls_mobile.onnx"
+_BUNDLED_RAPIDOCR_REC = "PP-OCRv6_rec_small.onnx"
```
jakub-walaszczyk: Any constants should be elevated to the top of the file. The question is, why do we have such relative paths here? Shouldn't we store such paths in the `pipelines-components` after all? Question, not assumption. Let's discuss on why

ai4rag/components/data/text_extraction.py
```diff
@@ -350,24 +406,148 @@ def _resolve_artifacts_path(explicit: str | None) -> Path | None:
...
+        return _ARTIFACTS_RAPIDOCR_CHINESE
+    return _ARTIFACTS_RAPIDOCR_ENGLISH
+
+
+def _try_resolve_wheel_rapidocr_model_paths() -> dict[str, str] | None:
+    """Return RapidOCR wheel model paths when the installed package still ships ONNX files."""
+    try:
+        import rapidocr
+    except ImportError:
+        return None
```
jakub-walaszczyk: is there a case where we do not install rapidocr? I do not think so. It is most likely installed with docling's deps (btw I do not see the added dependencies, so where does this come from?)

Additionally, when we catch error on import and we silently return None this is completely not informational. How are we supposed to debug it later in the prod when all we have are some logs.

ai4rag/components/data/text_extraction.py
```diff
@@ -37,6 +42,7 @@
 _logger.addHandler(handler)
 
 DOWNLOAD_MAX_THREADS = 8
+DEFAULT_OCR_LANG: tuple[str, ...] = ("english",)
```
jakub-walaszczyk, line 59: What are the languages and how is the language detection handled? What is the process?

jakub-walaszczyk (changes requested):
Proper PR name is missing.
There is no description in the PR, even though the template is prepared.
Some things are not described on whjy we make things the way we make. Maybe we should consider some quick demo before processing further to introduce some more context for the review.

Additionally, please provide some more information on what happens with the language handling. It is kind of limited here, hence a little bit error prone and misaligned with what we actually try to support?

ai4rag/components/data/constants.py
Medokins: Deleted

ai4rag/components/data/text_extraction.py
Medokins: I described the supported languages and how the language detection is handeled, but Im not sure if this is the correct place (inline) to describe it

ai4rag/components/data/text_extraction.py
Medokins: Added docstring describing each parameter

ai4rag/components/data/text_extraction.py
Medokins: You're right, changed it to DoclingExtractionConfig

ai4rag/components/data/text_extraction.py
Medokins: Changed it to instance of DoclingExtractionConfig dataclass

ai4rag/components/data/text_extraction.py
Medokins: Moved consts to top of the file. 

As for they why;
They're a mirror of Docling's own internal contract. In docling/models/stages/ocr/rapid_ocr_model.py, Docling hardcodes the exact same strings in its model-set tables. Additionally (and this is the main reason I've done this) docling doesnt fail fast,  when a model file is missing under artifacts_path, Docling only does this:
```
if not Path(model_path).exists():
    _log.warning(f"The provided model path {model_path} is not found.")
# ...then proceeds to construct RapidOCR(params=params) anyway
```

ai4rag/components/data/text_extraction.py
Medokins: Added more info when rapidocr is missing on import

jakub-walaszczyk (approved):

jakub-walaszczyk (approved):

PR #143, opened 2026-09-02
feat: change the keys for paths

MRGuziX:
## Description
Changed the file id to the path key so we will distinguish between files of the same name  but on a different path, and we can choose the correct file for the RAG feature 

## Motivation
[Why is this change needed?
](https://redhat.atlassian.net/browse/RHOAIENG-88695)
## Changes
Benchmark field rename: correct_answer_document_ids → correct_answer_document_keys in benchmark_data.py and test_data_loader.py; old field is now rejected (not deprecated)
- Property rename: BenchmarkData.document_ids → BenchmarkData.document_keys with non-empty validation (each question must reference at least one document key)
- evaluation_results.json: answer_contexts[].document_id → answer_contexts[].document_key in results.py and rag_templates_optimization.py
- Text extraction: documents are named by their relative S3 object key (e.g. manuals/xr-200/file.txt) instead of bare filename; input_data_key prefix is stripped cleanly via a new parameter instead of hardcoding "documents" as a path marker
- Document loading fix: load_docling_documents now recursively scans subdirectories (rglob) so documents saved in nested directory structures are found
- Validation: mps.py validates benchmark keys against ingested documents — fails clearly if any key is missing
- Tests updated: all unit, functional, and integration tests updated to use new field names; backward-compat tests replaced with rejection tests confirming old field is no longer accepted

## Testing
-dev image
-pipeline from newest PLC changes
-run experiment
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing


jakub-walaszczyk:
@MRGuziX there were some changes in `ai4rag` components / data. Please see those and make proper updates. Thanks and sorry for the trouble

ai4rag/core/experiment/benchmark_data.py
```diff
@@ -50,18 +50,25 @@ def __init__(self, benchmark_data: pd.DataFrame):
...
         self.questions: list[str] = list(self._benchmark_data[self.QUESTION])
         self.correct_answers: list[list[str]] = list(self._benchmark_data[self.CORRECT_ANSWERS])
-        self.document_ids: list[list[str]] = list(self._benchmark_data[self.DOC_IDS])
+
+        if self.DOC_KEYS not in self._benchmark_data.columns:
+            raise BenchmarkDataValueError(
+                f"Benchmark data must contain '{self.DOC_KEYS}'. "
+                "Each record needs a list of S3 object keys identifying the ground-truth documents."
+            )
+        self.document_keys: list[list[str]] = list(self._benchmark_data[self.DOC_KEYS])
```
jakub-walaszczyk, line 76: To run `ai4rag` user does not need to have s3 access. Data can be read from the local drive. Should we introduce such message in here? I think in the ai4rag context, the key should be treated as document identifier. Document identifier in the pure `ai4rag` context can be anything that uniquely identifies the file.

My proposal: remove reference to s3, use the `correct_answer_document_keys` as required.

ai4rag/core/experiment/benchmark_data.py
```diff
@@ -30,8 +30,8 @@ class BenchmarkData:
     correct_answers : list[str]
         Validated answers from the benchmark dataset.
 
-    document_ids : list[str]
-        Validated IDs of documents with correct context for given answers.
+    document_keys : list[list[str]]
+        Validated S3 object keys of documents with correct context for given answers.
```
jakub-walaszczyk: I think we can drop the direct s3 reference and introduce some statement that document_keys are unique document identifiers (relative paths when provided from local drive or actual keys when using s3)?

ai4rag/core/experiment/benchmark_data.py
```diff
@@ -41,7 +41,7 @@ class BenchmarkData:
 
     QUESTION = "question"
     CORRECT_ANSWERS = "correct_answers"
-    DOC_IDS = "correct_answer_document_ids"
+    DOC_KEYS = "correct_answer_document_keys"
```
jakub-walaszczyk, line 58: Using DOC_IDS here was technically correct. In the context of `ai4rag` these documents names, keys or w/e are treated as unique documents identifiers. Something to consider, both options are fine.

ai4rag/core/experiment/mps.py
```diff
@@ -148,11 +148,19 @@ def evaluate_patterns(self):
...
+        for element in self.benchmark_data.document_keys:
+            benchmark_keys.extend(element)
+
+        ingested_keys = {doc.name for doc in self.documents}
+        missing_keys = set(benchmark_keys) - ingested_keys
+        if missing_keys:
+            raise ValueError(
+                f"Benchmark data references {len(missing_keys)} document keys not ingested: "
+                f"{sorted(missing_keys)[:5]}{'...' if len(missing_keys) > 5 else ''}"
+            )
```
jakub-walaszczyk: If we get benchmark data here, and then we sample the documents based on the benchmark data, this check makes not much sense, right? One line below you get all the documents needed.

ai4rag/utils/data/text_extraction.py
```diff
@@ -175,6 +175,7 @@ def extract_text(  # pylint: disable=too-many-locals,too-many-arguments,too-many
     max_extraction_workers: int | None = None,
     docling_artifacts_path: str | None = None,
     docling_config: DoclingExtractionConfig | None = None,
+    input_data_key: str = "",
```
jakub-walaszczyk: Is it necessary? For what I understand here, the `documents` parameter, that is most likely the results of documents discovery, has all the description with documents keys. Why would we need `input_data_key` here? I believe the idea about documents validation regarding s3 is done in `documents_discovery`.

ai4rag/utils/data/text_extraction.py
```diff
@@ -800,9 +827,18 @@ def _download_and_submit(  # pylint: disable=too-many-locals
                 continue
             downloaded_paths.append(local_path)
 
+    # Map local paths back to their S3 keys for doc_name computation
+    local_to_key = {(download_path / d["key"]).resolve(): d["key"] for d in supported}
```
jakub-walaszczyk: Can't we e.g. keep the full keys in the `documents: list[dict]`?

jakub-walaszczyk (commented):
Please address the questions regarding ai4rag local usage. We need to tell user how to act with `ai4rag` locally with no s3.

Please review and update the notebooks accordingly if applicable.

Let's revisit the idea of coming with the `input_data_key` to text `extract_text()`.

ai4rag/core/experiment/benchmark_data.py
MRGuziX: You're right that either works. I kept DOC_KEYS only so the constant matches the column name it holds (correct_answer_document_keys) — the docstring now makes the "key == unique document identifier" meaning explicit, which was the real gap. Happy to switch back to DOC_IDS if you'd rather.

ai4rag/core/experiment/mps.py
MRGuziX: Fair point, the hard failure was too strict. Changed it to a warning that names the missing keys, since silently dropping ground-truth documents lowers the retrieval scores without any signal.

ai4rag/core/experiment/benchmark_data.py
MRGuziX: Done. The message no longer mentions S3 and now explains both case

ai4rag/utils/data/text_extraction.py
MRGuziX: fair point

ai4rag/utils/data/text_extraction.py
MRGuziX: You're right, it doesn't belong here.
I will adjust the PLC changes also

ai4rag/core/experiment/benchmark_data.py
MRGuziX: sure

ai4rag/core/experiment/mps.py
```diff
@@ -148,11 +148,29 @@ def evaluate_patterns(self):
...
+
+        missing_keys = benchmark_keys - {document.name for document in self.documents}
+        if missing_keys:
+            logger.warning(
+                "Benchmark data references %d document key(s) that were not ingested. "
+                "Those documents are skipped, which lowers the retrieval scores: %s%s",
+                len(missing_keys),
+                sorted(missing_keys)[:5],
+                "..." if len(missing_keys) > 5 else "",
+            )
```
jakub-walaszczyk: This should be removed. If you want to introduce such logic, AI4RAGExperiment would be a better place to validate whether there are document keys covering the benchmark_data requirements. Drop this validation logic from the mps.

BTW: if there are no grounding documents for some reason, the scores and models responses should reflect that

ai4rag/core/experiment/mps.py
```diff
@@ -148,11 +148,29 @@ def evaluate_patterns(self):
...
+
+        documents = [document for document in self.documents if document.name in benchmark_keys]
+        if not documents:
+            raise ValueError(
+                "None of the document keys referenced by the benchmark data match an ingested "
+                "document, so there is nothing to pre-select on. Check that "
+                f"'{BenchmarkData.DOC_KEYS}' holds ingested document names, e.g. "
+                f"{sorted(benchmark_keys)[:3]} vs. {sorted(d.name for d in self.documents)[:3]}."
+            )
```
jakub-walaszczyk: Same here. This validation can be taken out to the AI4RAGExperiment. Here we should not worry about data validation.

ai4rag/utils/data/documents_discovery.py
```diff
@@ -28,13 +28,25 @@ class DocumentDescriptor:
     Attributes
     ----------
     key : str
-        Full S3 object key.
+        Full S3 object key, used to fetch the object.
     size_bytes : int
         Object size in bytes.
+    relative_key : str
```
jakub-walaszczyk: Why do we need relative key at all?

MRGuziX: `datasets/rag/v1/docs/manuals/xr-200/setup.pdf` -> that is full s3 key
`manuals/xr-200/setup.pdf` -> that is relative key

on main(branch) a document's name was just the filename, so `docs/a/setup.txt` and `docs/b/setup.txt` both became setup.txt — one shadowed the other. relative_key gives each document a unique, path-aware name

jakub-walaszczyk: but full object key gives you that + more. relative_key seems to be redundant when you have just key.

jakub-walaszczyk (approved):

PR #145, opened 2026-09-04
refactor(vector_store)!: switch pgvector backend from psycopg to asyncpg

jakub-walaszczyk:
psycopg needs a libpq binding in every install mode: bundled (`[binary]`, which statically ships its own libpq/OpenSSL and can conflict with an OpenSSL already loaded by another dependency), compiled (`[c]`, needs a compiler and libpq headers at install time), or system-provided (plain `psycopg`, needs libpq installed on the target machine at runtime — the failure mode that motivated this change). asyncpg reimplements the Postgres wire protocol itself and uses Python's own `ssl` module, so it ships as a normal wheel with no libpq dependency in any form, matching ai4rag's requirement to install and run out of the box.

PGVectorStore's public sync API (search/add_documents/clean_collection/ close) is unchanged: asyncpg's Pool/Connection are bound to the event loop that creates them, so the store now opens one dedicated background event loop per instance and dispatches every DB coroutine onto it via run_coroutine_threadsafe, bridging back to a plain return value for every caller thread (e.g. query_rag()'s ThreadPoolExecutor workers).

Also fixes a latent ordering bug surfaced while rewriting _configure_connection: register_vector() looks up the `vector` type's OID, which does not exist until `CREATE EXTENSION IF NOT EXISTS vector` has run at least once, but the old code called them in the wrong order.

BREAKING CHANGE: PGVectorStore now raises asyncpg exceptions on database errors (e.g. asyncpg.exceptions.PostgresConnectionError) instead of psycopg ones. `psycopg`/`psycopg_pool` are no longer dependencies of ai4rag; `asyncpg` is required instead.

Assisted-by: Claude Code

## Description
Brief summary of changes

## Motivation
Why is this change needed?

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/rag/vector_store/pgvector.py
+++ b/ai4rag/rag/vector_store/pgvector.py
@@ -6,6 +6,7 @@ import asyncio
 import heapq
 import json
 import threading
+from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
 
@@ -37,6 +38,20 @@ class _LoopBoundPool:
     pool: asyncpg.Pool
     loop: asyncio.AbstractEventLoop
     thread: threading.Thread
+
+
+class _InFlightTracker:
+    """Tracks calls dispatched via :meth:`PGVectorStore._run`, so :meth:`PGVectorStore.close`
+    can wait for them to drain before tearing down the event loop.
+
+    ``cond`` and ``count`` are always read and mutated together — one under the
+    other — so, like :class:`_LoopBoundPool`, they are bundled into a single
+    object rather than two separate attributes on the store.
+    """
+
+    def __init__(self) -> None:
+        self.cond = threading.Condition()
+        self.count = 0
 
 
 class PGVectorStore(BaseVectorStore):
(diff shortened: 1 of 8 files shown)


jakub-walaszczyk:
@Mateusz-Switala notebooks work when downloaded to the local env. 
@MichalSteczko @Mateusz-Switala did you have a chance to look at this? Can we merge and release?

ai4rag/rag/vector_store/pgvector.py
```diff
@@ -148,112 +211,233 @@ def __init__(
...
         self._table_created: bool = False
         self._db_lock = threading.Lock()
 
-    def _ensure_pool(self) -> ConnectionPool:
-        """Return the pool, opening it on the first call — does not create the table.
+        # Guards the handoff between _run() (reader) and close() (writer): close()
+        # must not stop/close the loop while another thread's _run() call is still
+        # dispatched on it. See _run() and close() for the drain protocol this
+        # implements.
+        self._inflight = _InFlightTracker()
```
MichalSteczko, line 223: Add this attribute to __init__ and initialise it there since other callers can get "different" Conditions to stop or maybe I am mistaken and it is expected behaviour.

Mateusz-Switala (approved):
/lgtm

ai4rag/rag/vector_store/pgvector.py
```diff
@@ -148,112 +211,233 @@ def __init__(
...
+                coro.close()  # avoid a "coroutine was never awaited" warning
+                raise RuntimeError(f"PGVectorStore for collection {self._collection_name!r} is closed.")
+            self._inflight.count += 1
+        try:
+            return self._dispatch(db.loop, coro)
+        finally:
+            with self._inflight.cond:
+                self._inflight.count -= 1
+                if self._inflight.count == 0:
+                    self._inflight.cond.notify_all()
```
MichalSteczko: Do we need to spawn here another `condition` since it can't cause race condition.

ai4rag/rag/vector_store/pgvector.py
```diff
@@ -310,60 +494,75 @@ def _ensure_indexes(self) -> None:
...
             A ``CREATE INDEX IF NOT EXISTS ...`` statement.
         """
         try:
-            conn.execute(index_sql)
-        except psycopg.errors.UniqueViolation:
+            await conn.execute(index_sql)
+        except asyncpg.exceptions.UniqueViolationError:
             logger.info(
                 "PGVector index for %s was created concurrently elsewhere; continuing.",
                 self._collection_name,
```
MichalSteczko, line 568: Maybe we should to put this on DEBUG.

ai4rag/rag/vector_store/pgvector.py
```diff
@@ -736,19 +1002,48 @@ def _insert_batch_with_retry(self, batch: list[tuple[str, str, list[float], str,
...
             logger.warning("PGVector insert failed (%s); retrying batch of %d rows.", exc, len(batch))
-            self._insert_batch(batch)
+            self._run(self._insert_batch_async(pool, batch))
 
     def clean_collection(self) -> None:
         """Drop the PostgreSQL table."""
-        with self._ensure_pool().connection() as conn:
-            conn.execute(f"DROP TABLE IF EXISTS {self._quoted_table()} CASCADE")
+        pool = self._ensure_pool()
+        self._run_with_retry(lambda: self._drop_table(pool))
```
MichalSteczko, line 1018: `clean_collection()` method drops the table but does not reset `_table_created` and `_indexes_built`. Subsequent operations may assume the dropped table/indexes still exist.

MichalSteczko (commented):
LGTM

PR #150, opened 2026-09-10
feat: introduce warm start

Mateusz-Switala:
## Description

  Improves the warm-start strategies for the GAM-based HPO optimizer (`GAMOptimizer`) so greedy and balanced
  starts provide meaningful discrete-parameter coverage before GAM-guided search begins.

  ## Motivation

  The previous warm-start selection could under-cover discrete parameter values, causing early trials to
  cluster and leaving the fitted GAM with sparse categorical observations.

  ## Changes

  - **Coverage-aware warm starts:** Greedy and balanced strategies prioritize combinations that cover
  previously unseen discrete values.
  - **Balanced strategy:** Uses round-robin selection across `fields_to_balance` value tuples while also
  prioritizing coverage of non-balanced discrete fields.
  - **Automatic warm-start sizing:** For greedy and balanced strategies, the effective warm-start target
  expands to the minimum required for coverage rather than failing or silently under-covering when
  `n_random_nodes` is too small.
  - **Evaluation allocation:** `max_evals` is optional; when specified, output allocation accounts for the
  effective warm-start target and reserves remaining capacity for GAM-guided evaluations.
  - **Model-value handling:** Normalizes model objects, dictionaries, and strings for coverage tracking;
  serializes model-valued DataFrame columns to `model_id` before GAM fitting.
  - **Balanced-selection fix:** Corrected the ordering/fallback behavior so round-robin balancing is
  preserved.
  - **Tests:** Expanded `test_gam_opt.py` for coverage, round-robin behavior, model serialization, auto-
  adjusted warm starts, and post-warm-start GAM iterations.

  ## Testing

  - `tests/unit/ai4rag/core/hpo/test_gam_opt.py`

  ## Checklist

  - [x] Tests added/updated
  - [ ] Documentation updated
  - [ ] All checks passing


ai4rag/core/experiment/experiment.py
```diff
@@ -155,6 +155,8 @@ def __init__(
         self.inference_max_threads: int = kwargs.pop("inference_max_threads", 10)
 
         self.results: ExperimentResults = ExperimentResults()
+        self.optimizer: BaseOptimizer | None = None
```
filip-komarzyniec: Due to the way the `self.optimizer` is handled in the `search` function I don't think it makes sense to include it here. 

What's more, making arguments `| None`-able usually complicates things in linters-like tools resulting in lots of unnecessary warnings. 
(1)`None` value for optimizer makes little sense since it's completely required for the experiment to run. 
(2) We do not support providing the optimizer directly to the experiment class. The pattern used here is more of an Dependency Injection-like pattern where externally constructed optimiser is injected directly into search function.

ai4rag/core/experiment/experiment.py
```diff
@@ -685,6 +687,9 @@ def search(self, **kwargs) -> None:
         """
 
         logger.info("Starting RAG optimization process...")
+        # GAM patterns are buffered until the optimizer has completed, so clear
+        # results from a previous invocation on this experiment instance.
+        self._optimization_patterns = []
```
filip-komarzyniec: Is it even a valid scenario? We do not run `search` function anywhere multiple times. Usually the flow is: create experiment --> run search --> exit python process (no way to re-run search). 

If clear state at the beginning is necessary why not make it just a local variable?

ai4rag/core/experiment/experiment.py
```diff
@@ -823,10 +879,16 @@ def _stream_finished_pattern(
             "iteration": len(self.results) + n_known,
         }
 
-        self.event_handler.on_pattern_creation(
-            payload=payload,
-            evaluation_results=evaluation_results_json,
-        )
+        if isinstance(self.optimizer, GAMOptimizer):
```
filip-komarzyniec: Do we support any other optimiser? AFAIK not, let's not write unnecessarily generic code

ai4rag/core/experiment/experiment.py
```diff
@@ -731,17 +736,68 @@ def objective_function(space: RAGParamsType) -> float | None:
...
         )
 
+        self.optimizer = optimizer
         try:
             _ = optimizer.search()
         except OptimizationError as err:
             final_error_msg = self._exception_handler.get_final_error_msg()
             raise RAGExperimentError(final_error_msg) from err
 
+        self._publish_optimization_patterns()
```
filip-komarzyniec: This and `_select_optimization_patterns` introduce quite a mess. 

(1) they override and shadow the intended behaviour of `ExperimentResults` class in terms of choosing the best pattern. Cannot we reuse that? 
What about choosing the best patterns from `ExperimentResults` class? Will it work the same or will it return different patterns?

(2) they complicate and partially override the previous flow with `_stream_finished_pattern` method. The previously simple flow branching is quite complicated now and the same behaviour and responsibility (e.g. calling event_handler callbacks) spans multiple functions.

filip-komarzyniec (commented):
Some architectural doubts wrt. the Experiment class.

The GamOpt internals look conceptually okay and are covered with extensive unit tests so I take them as-is 🙋🏼‍♂️

filip-komarzyniec (approved):

piotrhm (approved):
All comments and remarks have been addressed privately. Approved.

ai4rag/core/experiment/experiment.py
Mateusz-Switala: Removed - I've refactored a bit to get rid of `_select_optimization_patterns` and now the `self.optimizer` is not needed.

ai4rag/core/experiment/experiment.py
Mateusz-Switala: Actually we do: https://github.com/IBM/ai4rag/blob/469932229ab41ec6d2d2f9f37dffdedfb5a5471b/ai4rag/core/hpo/random_opt.py#L30 , I think

ai4rag/core/experiment/experiment.py
Mateusz-Switala: `_select_optimization_patterns` has been removed entirely and using now xperimentResults. Thanks!

ai4rag/core/experiment/experiment.py
Mateusz-Switala: This has been already removed. We use now ExperimentResults as the single result store instead

LukaszCmielowski (approved):

PR #151, opened 2026-09-10
fix(search_space): log root cause when MaaS model validation fails

jakub-walaszczyk:
_validate_foundation_model/_validate_embedding_model and the embedding auto-detection failure path in _get_models caught the model-query exception and logged only the model id, discarding the underlying OpenAI client error entirely.

- add exc_info=True so the traceback is preserved in the warning log
- add _describe_error() to extract status_code/code/message from openai.APIError (and APIStatusError subclasses) when available, falling back to str(exc) otherwise

Assisted-by: Claude Code

## Description
Brief summary of changes

## Motivation
Why is this change needed?

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing

### DIFF
--- a/ai4rag/search_space/prepare/models.py
+++ b/ai4rag/search_space/prepare/models.py
@@ -39,7 +39,7 @@ from collections.abc import Mapping, Sequence
 from dataclasses import fields, is_dataclass
 from typing import Any
 
-from openai import OpenAI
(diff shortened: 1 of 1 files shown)


Mateusz-Switala (approved):
lgtm

PR #154, opened 2026-09-11
feat(vector_store)!: replace ChromaDB with Milvus Lite

jakub-walaszczyk:
Remove the ChromaDB backend entirely (its `chromadb` dependency carried a large vulnerability surface) and adopt Milvus Lite as the local, zero-server vector store. Milvus Lite is the embedded mode of the existing Milvus backend, so it reuses `MilvusVectorStore` and supports hybrid (dense + BM25) search that Chroma never had.

- Split the Milvus config in two, both served by `MilvusVectorStore`:
  - `MilvusConfig` (provider "milvus") is now remote-server-only and validates that `uri` is an `http(s)://` URL, so a mistyped `MILVUS_URI` fails loudly instead of silently spinning up a throwaway local database in production.
  - `MilvusLiteConfig` (provider "milvus_lite") is the explicit embedded engine, configured by a local `db_path` (env `MILVUS_LITE_DB_PATH`).
- Add `temporary_milvus_lite_store`, a throwaway-store context manager (temp dir + auto cleanup) now backing model pre-selection and judge calibration, which previously relied on ephemeral in-memory Chroma.
- Remove all Chroma vector-only special-casing: hybrid search and the full hybrid search space now apply to every backend; `vector_store_type` is validated against "milvus", "milvus_lite", and "pgvector".
- Swap the dependency `chromadb` -> `pymilvus[milvus-lite]` and delete the now-obsolete `ensure_sqlite3` ChromaDB compatibility shim.
- Update tests (real embedded Milvus Lite in the unit tier, plus integration and functional concurrent-query tests) and all documentation accordingly.

BREAKING CHANGE: `ChromaConfig`, `ChromaVectorStore`, and the "chroma" vector store provider are removed, along with the `chromadb` dependency. `MilvusConfig` no longer accepts a local file path as its `uri` (it must be an `http(s)://` server URL); use `MilvusLiteConfig(db_path=...)` for an embedded local database.

Assisted-by: Claude Code

Relates to: https://github.com/IBM/ai4rag/issues/153
Resolves:
- CVE-2026-45831
- CVE-2026-45830
- CVE-2026-45833

## Description
Brief summary of changes

## Motivation
Why is this change needed?

## Changes
- Another change

## Testing
How did you test this?

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] All checks passing


Mateusz-Switala (approved):
/lgtm

